PR #20, opened 2020-08-11
MC 2028 data and CRAB update

shedprog:
"Preliminary pull request"

For every dataset I'm not sure about or it is somehow different from others there is a comment above it "#"

PR #22, opened 2020-08-17
Production: adding Phase2 offline modifications to run out-of-the-box DeepTau

ArturAkh:
Dear all,

I've started to implement required modifications for Phase2 offline, such that it should be possible to run the production `cmsRun` configuration for out-of-the-box DeepTau for Phase2 & the config should remain compatible with all other up to now considered applications.

To achieve the latter point, a few further modifications would be needed, I guess, so I've got corresponding comments to the places with potential changes below.

Cheers,

Artur


Production/python/Production.py
```diff
@@ -85,13 +85,21 @@
     process.combinatoricRecoTaus.builders[0].signalConeSize = cms.string('max(min(0.2, 4.528/(pt()^0.8982)), 0.03)') ## change to quantile 0.95
     process.selectedPatTaus.cut = cms.string('pt > 18.')   ## remove DMFinding filter (was pt > 18. && tauID(\'decayModeFindingNewDMs\')> 0.5)
 
-tauIdConfig = importlib.import_module('TauMLTools.Production.runTauIdMVA')
+tauIdConfig = importlib.import_module('RecoTauTag.RecoTau.tools.runTauIdMVA')
```
ArturAkh, line 88: I've switched to the `runTauIdMVA` code from official CMSSW, since the one used within TauMLTools does not work with Phase2 currently. Is there any particular reason to stick the TauMLTools specific version of it (apart from better handling of debug & input tau collection)?

Production/python/Production.py
```diff
@@ -85,13 +85,21 @@
...
 updatedTauName = "slimmedTausNewID"
 tauIdEmbedder = tauIdConfig.TauIDEmbedder(
-    process, cms, debug = False, updatedTauName = updatedTauName,
-    toKeep = [ "2017v2", "dR0p32017v2", "newDM2017v2", "deepTau2017v2p1"]
+    process, cms, updatedTauName = updatedTauName,
+    toKeep = [ "2017v2", "dR0p32017v2", "newDM2017v2", "deepTau2017v2p1", "newDMPhase2v1"]
```
ArturAkh, line 92: I suppose I should add an `if else`  here to properly set the IDs to be kept for Phase2

Production/python/Production.py
```diff
@@ -85,13 +85,21 @@
     process.combinatoricRecoTaus.builders[0].signalConeSize = cms.string('max(min(0.2, 4.528/(pt()^0.8982)), 0.03)') ## change to quantile 0.95
     process.selectedPatTaus.cut = cms.string('pt > 18.')   ## remove DMFinding filter (was pt > 18. && tauID(\'decayModeFindingNewDMs\')> 0.5)
 
-tauIdConfig = importlib.import_module('TauMLTools.Production.runTauIdMVA')
+tauIdConfig = importlib.import_module('RecoTauTag.RecoTau.tools.runTauIdMVA')
 updatedTauName = "slimmedTausNewID"
 tauIdEmbedder = tauIdConfig.TauIDEmbedder(
-    process, cms, debug = False, updatedTauName = updatedTauName,
```
ArturAkh, line 91: Is the debug option needed for us?

Production/python/Production.py
```diff
@@ -85,13 +85,21 @@
...
-    toKeep = [ "2017v2", "dR0p32017v2", "newDM2017v2", "deepTau2017v2p1"]
+    process, cms, updatedTauName = updatedTauName,
+    toKeep = [ "2017v2", "dR0p32017v2", "newDM2017v2", "deepTau2017v2p1", "newDMPhase2v1"]
 )
-tauIdEmbedder.runTauID(tau_collection = tau_collection)
+tauIdEmbedder.runTauID()
```
ArturAkh, line 94: The official CMSSW version of `runTauIdMVA` does not have a possibility to modify the tau collection. Should we ask for a corresponding change in official CMSSW rather than using it the TauML specific one? You can find the (very nasty & not that bulletproof) workaround below...

ArturAkh:
I guess I also need to change the Geometry properly to the one from Phase2 here based on sampleType:

https://github.com/ArturAkh/TauMLTools/blob/master/Production/python/Production.py#L54-L55

However, do you think this has any influence, if there aren't any RECO sequences run? I suppose, that shouldn't be problematic as long as an analysis-like production is run, reading info only from MiniAOD.

Analysis/interface/TauIdResults.h
```diff
@@ -57,6 +57,8 @@ struct TauIdResults {
         "VVLoose VLoose Loose Medium Tight VTight VVTight") \
     TAU_ID(byIsolationMVArun2017v2DBoldDMdR0p3wLT2017, "by{wp}IsolationMVArun2017v2DBoldDMdR0p3wLT{raw}2017", true, \
         "VVLoose VLoose Loose Medium Tight VTight VVTight") \
+    TAU_ID(byIsolationMVADBnewDMwLTPhase2, "by{wp}IsolationMVADBnewDMwLTPhase2{raw}", true, \
+        "VVLoose VLoose Loose Medium Tight VTight VVTight") \
```
kandrosov, line 61: [FillTuple](https://github.com/cms-tau-pog/TauMLTools/blob/master/Analysis/interface/TauIdResults.h#L119-L134) should be modified to fill the raw and id branches with default values if ID not found. Otherwise, the code will not work for both Phase 2 and Run 2 samples.

Modified version (possibly some renaming will be needed) can be copied from [here](https://github.com/hh-italian-group/h-tautau/blob/master/Core/include/TauIdResults.h#L65-L82)

Production/crab/configs/Phase2_110X/DY.txt
```diff
@@ -0,0 +1,8 @@
+Automatic
+globalTag=110X_mcRun4_realistic_v3 sampleType=MC_Phase2_110X
```
kandrosov, line 2: Please, wait for #20 to be merged, because the format of crab txt files will slightly change.

Production/crab/configs/Phase2_110X/MuonGun.txt
```diff
@@ -0,0 +1,33 @@
+Automatic
+globalTag=110X_mcRun4_realistic_v3 sampleType=MC_Phase2_110X
+
+Muminus_Pt10-gun_NoPU    1000 /Muminus_Pt10-gun/Phase2HLTTDRWinter20RECOMiniAOD-NoPU_110X_mcRun4_realistic_v3-v2/MINIAODSIM
+Muminus_Pt10-gun_PU140   1000 /Muminus_Pt10-gun/Phase2HLTTDRWinter20RECOMiniAOD-PU140_110X_mcRun4_realistic_v3-v2/MINIAODSIM
+Muminus_Pt10-gun_PU200   1000 /Muminus_Pt10-gun/Phase2HLTTDRWinter20RECOMiniAOD-PU200_110X_mcRun4_realistic_v3-v2/MINIAODSIM
```
kandrosov, line 6: I don't think that muon gun with pt = 10 GeV will be useful for the training, because the cut on tau candidates stored in miniAOD is 18 GeV.

Production/crab/configs/Phase2_110X/QCD.txt
```diff
@@ -0,0 +1,8 @@
+Automatic
+globalTag=110X_mcRun4_realistic_v3 sampleType=MC_Phase2_110X
+
+QCD_Pt-15to3000_NoPU         1000 /QCD_Pt-15to3000_TuneCP5_Flat_14TeV-pythia8/Phase2HLTTDRWinter20RECOMiniAOD-NoPU_castor_all_pt_tracks_110X_mcRun4_realistic_v3-v2/MINIAODSIM
```
kandrosov, line 4: Are the any QCD samples with the correct tau spectrum available (in addition to the flat ones listed here)? The correct spectrum will be needed to reweight tau candidates when computing the ROC curves.

Production/crab/configs/Phase2_110X/Zprime.txt
```diff
@@ -0,0 +1,4 @@
+Automatic
+globalTag=110X_mcRun4_realistic_v3 sampleType=MC_Phase2_110X
+
+ZprimeToMuMu_M-6000_PU140 1000 /ZprimeToMuMu_M-6000_TuneCP5_14TeV-pythia8/Phase2HLTTDRWinter20RECOMiniAOD-PU140_110X_mcRun4_realistic_v3-v2/MINIAODSIM
```
kandrosov, line 4: There is no Z'->tautau or tau gun samples? Otherwise, the statistics for real tau with high pt will be very limited.

Production/python/Production.py
kandrosov: @lguzzi created a copy of the official `runTauIdMVA` to be able to re-run DeepTau after re-running tau reconstruction. I suggest to put all tau-id-related lines here inside `if else` for phase 2/run 2. This would also solve the two issues commented below.

Production/python/Production.py
```diff
@@ -85,13 +85,21 @@
...
+
+for objname in dir(process):
+    obj = getattr(process, objname)
+    if hasattr(obj, "PATTauProducer"):
+        setattr(obj,"PATTauProducer", cms.InputTag(tau_collection))
+    elif hasattr(obj, "taus"):
+        setattr(obj,"taus", cms.InputTag(tau_collection))
```
kandrosov, line 101: this is a too dirty hack, better to use if/else as suggested above :)

Production/python/sampleConfig.py
```diff
@@ -7,8 +7,9 @@
 from Configuration.Eras.Era_Run2_2016_cff import Run2_2016
 from Configuration.Eras.Era_Run2_2017_cff import Run2_2017
 from Configuration.Eras.Era_Run2_2018_cff import Run2_2018
+from Configuration.Eras.Era_Phase2C9_cff import Phase2C9
```
kandrosov, line 10: will this work with 10_6_X? If not need to implement some workaround. For example use `if else` inside GetPeriodCfg and do an `import` only inside the corresponding `if` branch.

kandrosov:
> I guess I also need to change the Geometry properly to the one from Phase2 here based on sampleType:
> 
> https://github.com/ArturAkh/TauMLTools/blob/master/Production/python/Production.py#L54-L55
> 
> However, do you think this has any influence, if there aren't any RECO sequences run? I suppose, that shouldn't be problematic as long as an analysis-like production is run, reading info only from MiniAOD.

we plan to rerun tau reco on miniAOD for Run 2 production. Maybe it would be safer to put these two lines inside `if else`. On the other hand, if the global tag is set correctly, I don't know if importing geometry can make any influence.

Production/crab/configs/Phase2_110X/DY.txt
ArturAkh: Ok, that's fine with me. I'll rebase & update Phase2 lists as soon as #20 is merged.

Production/crab/configs/Phase2_110X/MuonGun.txt
ArturAkh: Ok. Then I'll remove these samples from the list.

Production/crab/configs/Phase2_110X/QCD.txt
ArturAkh: What was the name syntax for Run2 samples with correct tau spectrum? In any case, the samples I've added above are the only ones available at the moment, see e.g. [here](https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset+status%3D*+dataset%3D%2FQCD*%2F*Run4*%2FMINIAODSIM)

Production/crab/configs/Phase2_110X/Zprime.txt
ArturAkh: Nope, see [here](https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset+status%3D*+dataset%3D%2FZprime*%2F*Run4*%2FMINIAODSIM)

ArturAkh: I'll have a look for TauGun, though

Production/python/Production.py
ArturAkh: Ok :)

Production/python/sampleConfig.py
ArturAkh: I'll add a cmssw version specific `if else` then

Production/crab/configs/Phase2_110X/Zprime.txt
ArturAkh: Here is the full list of the considered [Phase2 campaign](https://cmsweb.cern.ch/das/request?view=list&limit=150&instance=prod%2Fglobal&input=dataset+status%3D*+dataset%3D%2F*%2FPhase2*Run4*%2FMINIAODSIM)

The only ones that seem to be fine are [MultiTau samples](https://cmsweb.cern.ch/das/request?view=list&limit=150&instance=prod%2Fglobal&input=dataset+status%3D*+dataset%3D%2FMultiTau*%2FPhase2*Run4*%2FMINIAODSIM)

ArturAkh: I think it should be fine though just to request the samples from the Tau POG side

Production/crab/configs/Phase2_110X/QCD.txt
kandrosov: For Run 2 there are pt-binned QCD samples, which are not available for Phase 2. However, maybe it is possible to use MinBias [samples](https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset+dataset%3D%2FMinBias*%2F*Run4*%2FMINIAODSIM) that should have a more physical spectrum.

Production/crab/configs/Phase2_110X/Zprime.txt
kandrosov: Yes, I think multi-tau samples should be for the training.
N.B. maybe in the future we'll need to add non-overlap criterion for shuffle&merge step: e.g. exclude taus if their isolation cones overlap from the default training.

Production/crab/configs/Phase2_110X/MuonGun.txt
ArturAkh: Done

Production/crab/configs/Phase2_110X/QCD.txt
ArturAkh: Done

Production/crab/configs/Phase2_110X/Zprime.txt
ArturAkh: Added MultiTau samples

Analysis/interface/TauIdResults.h
ArturAkh: Done & tested under `CMSSW_10_6_15` for DYJets MC

Production/python/Production.py
ArturAkh: Done

Production/python/Production.py
ArturAkh: solved

Production/python/Production.py
ArturAkh: solved

Production/python/Production.py
ArturAkh: solved

Production/python/Production.py
ArturAkh: Done

Production/python/sampleConfig.py
ArturAkh: Done

Production/python/Production.py
```diff
@@ -51,7 +55,11 @@
 process.MessageLogger.cerr.FwkReport.reportEvery = 100
 
 process.load('Configuration.StandardSequences.MagneticField_cff')
-process.load('Configuration.Geometry.GeometryRecoDB_cff')
+# include Phase2 specific configuration only after 11_1_X
+if int(cmssw_release_numbers[0]) >= 11 and int(cmssw_release_numbers[1]) >= 1:
```
kandrosov, line 59: Better to define a boolean `isPhase2 = int(cmssw_release_numbers[0]) >= 11 and int(cmssw_release_numbers[1]) >= 1` and use it here and in L97. It will make the condition more transparent and easier to adjust in the future, if needed.

Production/python/sampleConfig.py
```diff
@@ -3,11 +3,16 @@
...
 import FWCore.ParameterSet.Config as cms
+import os
 
 from Configuration.Eras.Era_Run2_2016_cff import Run2_2016
 from Configuration.Eras.Era_Run2_2017_cff import Run2_2017
 from Configuration.Eras.Era_Run2_2018_cff import Run2_2018
-from Configuration.Eras.Era_Phase2C9_cff import Phase2C9
+
+# include Phase2 specific configuration only after 11_1_X
+cmssw_release_numbers = os.environ.get('CMSSW_VERSION').replace('CMSSW_','').split("_")
```
kandrosov, line 13: I think it would be better if we avoid using CMSSW version in sampleConfig in order to simplify the long-term support (e.g. UL, Run3, other Phase2 geometries etc.).
What I meant before is to modify GetPeriodCfg in the following way:
```python
def GetPeriodCfg(sampleType):
    period = GetPeriod(sampleType)
    if period == 'Run2016':
        from Configuration.Eras.Era_Run2_2016_cff import Run2_2016
        return Run2_2016
    elif period == 'Run2017':
        from Configuration.Eras.Era_Run2_2017_cff import Run2_2017
        return Run2_2017
    elif period == 'Run2018':
        from Configuration.Eras.Era_Run2_2018_cff import Run2_2018
        return Run2_2018
    elif period == 'Phase2':
        from Configuration.Eras.Era_Phase2C9_cff import Phase2C9
        return Phase2C9
    else:
        raise RuntimeError('Period = "{}" is not supported.'.format(period))
```

Production/python/Production.py
ArturAkh: Ok, is now updated

Production/python/sampleConfig.py
ArturAkh: Ok, adjusted the method as you've suggested. Sorry for misunderstanding :)

Issue #15, opened 2020-07-28
TauTuple: store phase 2 specific information


kandrosov:
We should consider adding information from the following sources:
- Timing layer
- HGCAL

PR #25, opened 2020-08-31
Extend information for DNN input (general & Phase 2)

ArturAkh:
Dear all,

I've prepared an extension of input quantities for the DNN, which comprises general inputs & Phase 2 specific quantities:

- Electrons: shower shape variables (general), HGCal variables (Phase 2 specific), further minor extensions
- Muons: adding reconstruction type of the muon (both relevant for Run 3 and Phase 2)
- Isolated Tracks: extended n_hits information from hit patterns (both Run 3 and Phase 2)
- packed PFCandidates: timing information (Phase 2) and further minor extensions
- PV: timing information (Phase 2) & position uncertainty

Please note, that so far, the Data Formats for Phase 2 and Run 3 are identical, such that the no separation of Phase 2 specific quantities is needed in the first place. But, to save space, a switch would be useful, as suggested by Konstantin in the corresponding issue #15.

For the time-being, I'll keep the custom Electron Merger to avoid additional merge-topics, but as soon as the Anti-Electron PR is merged into official CMSSW, I'll remove it & switch to the templated one.

Other than that, I consider the list of quantities as complete for Phase 2, at least to my current knowledge after having a look at PAT & RECO formats of the objects of interest.

This means that in principle that code can be merged in case we would like to start the production fast, but I will work on further developments (like `TreeStorageMode`) in the next days.

Don't hesitate to share your thoughts on this with me :) 

Cheers,

Artur


ArturAkh:
In case of Muons, I still would need to update & test the MuonHitMatch class for Phase 2:
https://github.com/cms-tau-pog/TauMLTools/blob/master/Production/interface/MuonHitMatch.h

ArturAkh:
> In case of Muons, I still would need to update & test the MuonHitMatch class for Phase 2:
> https://github.com/cms-tau-pog/TauMLTools/blob/master/Production/interface/MuonHitMatch.h
> https://github.com/cms-tau-pog/TauMLTools/blob/master/Production/src/MuonHitMatch.cc

Extension is implemented & since formats in 10_6_X are the same, it runs also for 2018 samples without crashing

ArturAkh:
I've had a closer look at the following quantities that are always 0:
https://github.com/ArturAkh/TauMLTools/blob/master/Production/plugins/TauTupleProducer.cc#L671-L694

It seems like these are just dummy methods, at least for the time-being:
https://github.com/cms-sw/cmssw/search?q=numberOfTimingHits&unscoped_q=numberOfTimingHits
https://github.com/cms-sw/cmssw/search?q=numberOfBadHits&unscoped_q=numberOfBadHits
https://github.com/cms-sw/cmssw/search?q=numberOfMuonHits&unscoped_q=numberOfMuonHits

In contrast to that, other methods appear much more often:
https://github.com/cms-sw/cmssw/search?q=numberOfAllHits&unscoped_q=numberOfAllHits
https://github.com/cms-sw/cmssw/search?q=numberOfValidHits&unscoped_q=numberOfValidHits

(Although numberOfInactiveHits apprears also only twice, but it works)


Nevertheless, the n_hits & n_matches work for the various muon stations, as implemented in the MuonHitMatch files.
@kandrosov, should I have a closer look at these quantities before making the merge (also in context of Run3, since also numberOfMuonHits is always 0, for example)?

In any case, I've made a post in the tracking hypernews, let's see what comes out.

Cheers,

Artur

ArturAkh:
Ok the hypernews entry for the discussion of the quantities mentioned above can be found here:

https://hypernews.cern.ch/HyperNews/CMS/get/recoTracking/1886/1/1/1/1.html

It seems like this is still in development and needs some further effort to be implemented for isolatedTracks from MiniAOD. I'll suggest to keep these quantities then as they are in the code and update the proper readout of them later.

I consider the current status of the pull request as ready for review

Cheers,

Artur

Analysis/interface/TauTuple.h
```diff
@@ -267,7 +315,21 @@
...
+    TRACK_VAR4(Int_t, n_MuonHits, n_ValidMuonHits, n_LostMuonHits, n_BadMuonHits) /* number Of Muon Hits (all, valid, lost, bad) */ \
+    TRACK_VAR3(Int_t, n_ValidMuonDTHits, n_LostMuonDTHits, n_BadMuonDTHits) /* number Of Muon DT Hits (valid, lost, bad) */ \
+    TRACK_VAR3(Int_t, n_ValidMuonCSCHits, n_LostMuonCSCHits, n_BadMuonCSCHits) /* number Of Muon CSC Hits (valid, lost, bad) */ \
+    TRACK_VAR3(Int_t, n_ValidMuonRPCHits, n_LostMuonRPCHits, n_BadMuonRPCHits) /* number Of Muon RPC Hits (valid, lost, bad) */ \
+    TRACK_VAR3(Int_t, n_ValidMuonGEMHits, n_LostMuonGEMHits, n_BadMuonGEMHits) /* number Of Muon GEM Hits (valid, lost, bad) */ \
+    TRACK_VAR3(Int_t, n_ValidMuonME0Hits, n_LostMuonME0Hits, n_BadMuonME0Hits) /* number Of Muon ME0 Hits (valid, lost, bad) */ \
     TRACK_VAR(Int_t, n_InactiveHits) /* number Of Inactive Hits */ \
+    TRACK_VAR3(Int_t, n_AllHits_0, \
+                      n_AllHits_1, \
+                      n_AllHits_2) /* number Of All Hits
```
kandrosov, line 329: For `n_AllHits`, `n_LostHits`, `n_LostPixelHits` and `n_LostStripHits` better to have more explicit names instead of 0,1,2. E.g. `n_AllHits_TRACK`, `n_AllHits_MISSING_INNER` etc.

Production/plugins/TauTupleProducer.cc
```diff
@@ -483,15 +498,95 @@ class TauTupleProducer : public edm::EDAnalyzer {
...
             tauTuple().ele_deltaPhiEleClusterTrackAtCalo.push_back(ele->deltaPhiEleClusterTrackAtCalo());
             tauTuple().ele_deltaPhiSuperClusterTrackAtVtx.push_back(ele->deltaPhiSuperClusterTrackAtVtx());
             tauTuple().ele_deltaPhiSeedClusterTrackAtCalo.push_back(ele->deltaPhiSeedClusterTrackAtCalo());
 
-            tauTuple().ele_mvaInput_earlyBrem.push_back(ele->mvaInput().earlyBrem);
-            tauTuple().ele_mvaInput_lateBrem.push_back(ele->mvaInput().lateBrem);
-            tauTuple().ele_mvaInput_sigmaEtaEta.push_back(ele->mvaInput().sigmaEtaEta);
-            tauTuple().ele_mvaInput_hadEnergy.push_back(ele->mvaInput().hadEnergy);
-            tauTuple().ele_mvaInput_deltaEta.push_back(ele->mvaInput().deltaEta);
+            tauTuple().ele_mvaInput_earlyBrem.push_back(!ele->hasUserFloat("hgcElectronID:sigmaUU") ? ele->mvaInput().earlyBrem : default_int_value);
```
kandrosov, line 506: - Here and below `ele->hasUserFloat("hgcElectronID:sigmaUU")` is used as an indicator whatever electron is HGCAL or not. Better to define explicit boolean `const bool isHGCAL = ele->hasUserFloat("hgcElectronID:sigmaUU")` and use it in the conditions.

Production/plugins/TauTupleProducer.cc
```diff
@@ -483,15 +498,95 @@ class TauTupleProducer : public edm::EDAnalyzer {
...
+            tauTuple().ele_mvaInput_lateBrem.push_back(!ele->hasUserFloat("hgcElectronID:sigmaUU") ? ele->mvaInput().lateBrem : default_int_value);
+            tauTuple().ele_mvaInput_sigmaEtaEta.push_back(!ele->hasUserFloat("hgcElectronID:sigmaUU") ? ele->mvaInput().sigmaEtaEta : default_value);
+            tauTuple().ele_mvaInput_hadEnergy.push_back(!ele->hasUserFloat("hgcElectronID:sigmaUU") ? ele->mvaInput().hadEnergy : default_value);
+            tauTuple().ele_mvaInput_deltaEta.push_back(!ele->hasUserFloat("hgcElectronID:sigmaUU") ? ele->mvaInput().deltaEta : default_value);
+
+            // shower shape variables are available for non-HGCal electrons with pt > 5
+            tauTuple().ele_sigmaIetaIphi.push_back(!ele->hasUserFloat("hgcElectronID:sigmaUU") && ele->polarP4().pt() > 5 ? ele->sigmaIetaIphi() : default_value);
```
kandrosov, line 513: Similarly here, instead of repeating the pt cut value in each condition define `const bool hasShapeVars = !isHGCAL && ele->polarP4().pt() > 5;`

Production/plugins/TauTupleProducer.cc
```diff
@@ -483,15 +498,95 @@ class TauTupleProducer : public edm::EDAnalyzer {
...
+            tauTuple().ele_full5x5_eBottom.push_back(!ele->hasUserFloat("hgcElectronID:sigmaUU") && ele->polarP4().pt() > 5 ? ele->full5x5_eBottom() : default_value);
+            tauTuple().ele_full5x5_e2x5Left.push_back(!ele->hasUserFloat("hgcElectronID:sigmaUU") && ele->polarP4().pt() > 5 ? ele->full5x5_e2x5Left() : default_value);
+            tauTuple().ele_full5x5_e2x5Right.push_back(!ele->hasUserFloat("hgcElectronID:sigmaUU") && ele->polarP4().pt() > 5 ? ele->full5x5_e2x5Right() : default_value);
+            tauTuple().ele_full5x5_e2x5Top.push_back(!ele->hasUserFloat("hgcElectronID:sigmaUU") && ele->polarP4().pt() > 5 ? ele->full5x5_e2x5Top() : default_value);
+            tauTuple().ele_full5x5_e2x5Bottom.push_back(!ele->hasUserFloat("hgcElectronID:sigmaUU") && ele->polarP4().pt() > 5 ? ele->full5x5_e2x5Bottom() : default_value);
+
+            // Only phase2 electrons with !ele->isEB() obtain a value != default_{int_,}value via hasUserFloat() decision
+            tauTuple().ele_hgcal_sigmaUU.push_back(ele->hasUserFloat("hgcElectronID:sigmaUU") ? ele->userFloat("hgcElectronID:sigmaUU") : default_value);
```
kandrosov, line 552: I suggest defining a lambda function to simplify the code below:
```c++
const auto fillUserFloat = [&ele](std::vector<float>& output, const std::string& name) {
    output.push_back(ele->hasUserFloat("hgcElectronID:sigmaVV") ? ele->userFloat("hgcElectronID:sigmaVV") : default_value);
};

fillUserFloat(tauTuple().ele_hgcal_sigmaUU, "hgcElectronID:sigmaUU");
...
```

Production/src/MuonHitMatch.cc
```diff
@@ -72,6 +72,12 @@ void MuonHitMatch::CountHits(const pat::Muon& muon, CountMap& n_hits)
                         muon_n_hits = &n_hits.at(MuonSubdetId::CSC);
                     else if(hit_pattern.muonRPCHitFilter(hit_id))
                         muon_n_hits = &n_hits.at(MuonSubdetId::RPC);
+                    else if(hit_pattern.muonGEMHitFilter(hit_id))
+                        muon_n_hits = &n_hits.at(MuonSubdetId::GEM);
+                    else if(hit_pattern.muonME0HitFilter(hit_id)) {
+                        muon_n_hits = &n_hits.at(MuonSubdetId::ME0);
+                        station_index = GetStationIndex(1, false);
```
kandrosov, line 79: could you, please, add some comment here to document why this fix is needed for ME0?

kandrosov (changes requested):

Production/src/MuonHitMatch.cc
ArturAkh: Done

Production/plugins/TauTupleProducer.cc
ArturAkh: Done

Production/plugins/TauTupleProducer.cc
ArturAkh: Done

Production/src/MuonHitMatch.cc
kandrosov: Just for my understanding, why `hit_pattern.getMuonStation(hit_id)` fails to return the correct station for ME0 and works for other sub-detectors? Is it a known bug in CMSSW? Instead, `segment.station()` works fine when counting matches for ME0 (in `CountMatches`)?

Production/plugins/TauTupleProducer.cc
```diff
@@ -483,15 +498,98 @@ class TauTupleProducer : public edm::EDAnalyzer {
...
+            const bool isHGCAL = ele->hasUserFloat("hgcElectronID:sigmaUU");
+            tauTuple().ele_mvaInput_earlyBrem.push_back(!isHGCAL ? ele->mvaInput().earlyBrem : default_int_value);
+            tauTuple().ele_mvaInput_lateBrem.push_back(!isHGCAL ? ele->mvaInput().lateBrem : default_int_value);
+            tauTuple().ele_mvaInput_sigmaEtaEta.push_back(!isHGCAL ? ele->mvaInput().sigmaEtaEta : default_value);
+            tauTuple().ele_mvaInput_hadEnergy.push_back(!isHGCAL ? ele->mvaInput().hadEnergy : default_value);
+            tauTuple().ele_mvaInput_deltaEta.push_back(!isHGCAL ? ele->mvaInput().deltaEta : default_value);
+
+            // shower shape variables are available for non-HGCal electrons with pt > 5
+            const bool hasShapeVars = !isHGCAL && ele->polarP4().pt() > 5;
+            tauTuple().ele_sigmaIetaIphi.push_back(!hasShapeVars ? ele->sigmaIetaIphi() : default_value);
```
kandrosov, line 516: here and below: conditions are inverted - should be `hasShapeVars ? ... : default_value`

ArturAkh: Right sorry ;) Fixed that now.

Production/src/MuonHitMatch.cc
ArturAkh: This is something I've just figured out by having a look at the outcome of these functions & what is then stored in the n_hits and n_matches quantities. Would need to ask & cross-check that on hypernews (I think tracking POG), why this is the case, to be honest.

I guess this is the case, since the tracking information (also for muon systems) seems to be still in development for Phase2, such that the new muonStations don't work properly, at least ME0

Production/plugins/TauTupleProducer.cc
ArturAkh: Done

Analysis/interface/TauTuple.h
ArturAkh: Done

ArturAkh:
Ok I guess so far, all changes requested up to now are implemented.

I think I should rebase on the current status of the Tau POG master, first?

Production/src/MuonHitMatch.cc
kandrosov: Ok, thanks for the details! For this PR, let's keep the workaround as you implemented it. Then, after we have feedback from the tracking POG, we might come back to it.

kandrosov:
> Ok I guess so far, all changes requested up to now are implemented.
> 
> I think I should rebase on the current status of the Tau POG master, first?

Thank you! There are no conflicts => no need to rebase. I'm going to merge the PR.

kandrosov (approved):

ArturAkh:
Ok thanks!

PR #26, opened 2020-09-02
Adding sampleType to the tuple variables and fixing Producer.py to run on data

lguzzi:
adding sample type to the tuple variables: filling data, MC or embedded.
Also fixing Producer.py not running on data samples

### DIFF
--- a/Production/python/Production.py
+++ b/Production/python/Production.py
@@ -91,6 +91,10 @@ if options.rerunTauReco:
     tauAtMiniTools.addTauReReco(process)
     tauAtMiniTools.adaptTauToMiniAODReReco(process, options.reclusterJets)
 
+    if isData:
+        from PhysicsTools.PatAlgos.tools.coreTools import runOnData
+        runOnData(process, names = ['Taus'], outputModules = [])
+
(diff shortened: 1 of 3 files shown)


Production/python/Production.py
```diff
@@ -118,7 +123,8 @@
 objectdR = 0.5
 
 process.tauTupleProducer = cms.EDAnalyzer('TauTupleProducer',
-    isMC                            = cms.bool(not isData),
+    isMC                            = cms.bool(not isData if not isEmbedded else False),
+    isEmbedded                      = cms.bool(not isData if     isEmbedded else False),
```
kandrosov, line 127: - Embedded should have `isMC` set to `True` in order to fill gen info => `isMC = cms.bool(not isData)`
- `isEmbedded = cms.bool(isEmbedded)`

kandrosov (changes requested):

Production/python/Production.py
lguzzi: done baf22d5e70aa0cd424b3b7603e5a25036bb07361

kandrosov (approved):

PR #29, opened 2020-09-18
DeepTau training for Phase-2 HLT

veelken:
Hi, 

please do NOT merge this PR.
I have created the PR, so that you can see which files I have changed for the Phase-2 HLT, because this makes it easiest to discuss how best to proceed.

There are a few things, which may be problematic:
1) The definition of the sample type in line 117 in  Analysis/bin/CreateBinnedTuples.cxx 
2) the definition of tau ID discriminators in Analysis/interface/TauIdResults.h 
3) the definition of tau ID discriminators in Training/python/common.py 

Konstantin, could you have a look at the changes, please, and let me know if you spot other things that would be problematic to merge and how best to proceed with the items 1-3 above ?

Thank you,

Christian


kandrosov:
Modifications in this PR are done on top of the old DeepTau v2 code structure. Current mainstream developments have considerably diverged from that and back-compatibility with the old FW version is not foreseen. I'm going to merge this PR into a separate branch to preserve the code used for Phase2 HLT DeepTau training.

PR #30, opened 2020-10-07
Shuffling and merge 

shedprog:
It is the beginning of the work on shuffling and merge procedure. Below, there are scripts for creating eta/pt histograms per dataset. It was tasted only on a single dataset "W2JetsToLNu_TuneCP5_13TeV-madgraphMLM-pythia8". And for 285 files (46Gb) what corresponds to ~2E+07 taus it takes ~30-40 mins, so it is per mean-size dataset what is not satisfying in my understanding (we have 86 datasets in sum) and that is just to produce spectrum....

1) I also tried hadd + Project, but hadd takes even more time than CreateSpectralHists
2) Should some htcondor script be written for this?


Analysis/bin/CreateSpectralHists.cxx
```diff
@@ -0,0 +1,145 @@
...
+    using TauTuple = tau_tuple::TauTuple;
+
+    CreateSpectralHists(const Arguments& args) :
+      input_files(RootFilesMerger::FindInputFiles(args.input_dirs(),
+                                                  args.file_name_pattern(),
+                                                  args.exclude_list(),
+                                                  args.exclude_dir_list())),
+       outputfile(root_ext::CreateRootFile(args.outputfile()))
+    {
+      auto par_path = GetPathWithoutFileName(args.outputfile());
```
kandrosov, line 71: To speed-up the decompression during reading (which takes most of the time), you can enable an implicit multithreading in ROOT:
```c++
ROOT::EnableThreadSafety();
if(args.n_threads() > 1)
    ROOT::EnableImplicitMT(args.n_threads());
```

Analysis/bin/CreateSpectralHists.cxx
```diff
@@ -0,0 +1,145 @@
...
+    eta_bins = analysis::Parse<int>(args_eta[0]);
+    eta_min = analysis::Parse<double>(args_eta[1]);
+    eta_max = analysis::Parse<double>(args_eta[2]);
+  }
+};
+
+class HistSpectrum : public root_ext::AnalyzerData {
+
+public:
+  HistSpectrum(std::shared_ptr<TFile> outputFile, HistArgs hargs) : root_ext::AnalyzerData(outputFile)
```
kandrosov, line 41: `const HistArgs& hargs`

Analysis/bin/CreateSpectralHists.cxx
```diff
@@ -0,0 +1,145 @@
...
+             ParseHistSetup(args.pt_hist(),args.eta_hist()));
+    }
+
+    void Run()
+    {
+        for(const auto& file_name : input_files) {
+            std::cout << "file: " << file_name << std::endl;
+            auto file = root_ext::OpenRootFile(file_name);
+
+            TauTuple input_tauTuple("taus", file.get(), true);
```
kandrosov, line 85: Reading all branches is very slow, enable only branches that you need (e.g. pt, eta, tau_type, tau_valid...). The list of enabled branches can be passed to the TauTuple constructor.

Analysis/bin/CreateSpectralHists.cxx
```diff
@@ -0,0 +1,145 @@
...
+            for(const Tau& tau : input_tauTuple)
+            {
+              AddTau(tau);
```
kandrosov, line 88: Add only valid taus.

Analysis/python/CreateSpectralHists.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python
```
kandrosov, line 1: I don't think that we want to run multiple instances of `CreateSpectralHists` at the same machine in parallel. This would also require some mechanism to track jobs which are failed (for one reason or another). I propose to keep things simple for the moment: run CreateSpectralHists consecutively one folder at the time and use implicit multithreading for reading ROOT file (see comment above). If it doesn't work sufficiently fast, we can think about more elaborate parallelization ways.

kandrosov (changes requested):

Analysis/bin/CreateSpectralHists.cxx
shedprog: @kandrosov Konstantin, do I understand correctly that by valid Taus we mean the following 
[](https://github.com/cms-tau-pog/TauMLTools/blob/master/Analysis/bin/CreateBinnedTuples.cxx#L149-L158)

shedprog: @kandrosov do I understand correctly that by tau_valid we mean the following criteria:

https://github.com/cms-tau-pog/TauMLTools/blob/c4f2ac5389198b864a0062d6830cee5cd0b13eaf/Analysis/bin/CreateBinnedTuples.cxx#L149-L158

kandrosov: I think we can relax these conditions and only require `tau_index >= 0`.

Analysis/bin/CreateSpectralHists.cxx
shedprog: @kandrosov enabling only branches that I need reduced time from 40mins to ~5-6mins, but for some reasons ROOT::EnableImplicitMT with more than one thread slows down everything significantly,

kandrosov: Interesting. Have you tested it several times (or on a machine where you fully control the conditions)? For lxplus, for example, from one run to another it can happen that someone else started a heavy job and everything goes slower...

However, it can be the case that the amount of enabled branches is so small, so there is no gain from parallelizing the decompression.

Please, choose the fastest option. Below 10 minutes per a large dataset (as W2JetsToLNu) is acceptable - we need to create spectrum histograms only once per production.

shedprog: Yes, I've tested it many times, the time increases very significantly, I am testing it on desy naf-cms.desy.de the machine is not loaded at all, if I run let's say with --n-threads 6 it loads full 6 cores

kandrosov: Ok, thanks for the details! I propose then to keep running the code with a single thread and move forward.
Creating binned spectrum is not a recurrent task: it enough to create histograms with fine binning (e.g. 1 GeV step for pt and 0.01 for eta) for each dataset once and then rebin them according to the needs when running ShuffleMerge. Therefore, as long as the total time to produce a spectrum for the all datasets is O(1 day) - it is acceptable.

Maybe you can also add histograms with the pt-eta spectrum of seeding jets - it is not needed for the current training, but can be used in the future (for example, for an ML-based tau reco training).

Analysis/bin/CreateSpectralHists.cxx
```diff
@@ -0,0 +1,161 @@
...
+    run::Argument<std::string> exclude_list{"exclude-list", "comma separated list of files to exclude", ""};
+    run::Argument<std::string> exclude_dir_list{"exclude-dir-list",
+                                               "comma separated list of directories to exclude", ""};
+    run::Argument<unsigned> n_threads{"n-threads", "number of threads", 1};
+};
+
+struct HistArgs {
+  int eta_bins, pt_bins;
+  double eta_min, eta_max, pt_min, pt_max;
+  HistArgs(std::vector<std::string> args_pt, std::vector<std::string> args_eta)
```
kandrosov, line 28: `const std::vector<std::string>&`

Analysis/bin/CreateSpectralHists.cxx
```diff
@@ -0,0 +1,161 @@
...
+  {
+    pt_bins = analysis::Parse<int>(args_pt[0]);
```
kandrosov, line 30: in general, use of `.at` is safer than `[]`

Analysis/config/2018/training_inputs_test.cfg
```diff
@@ -0,0 +1 @@
+jet_test: dir=(DY.*) file=eventTuple.* types=jet,e,mu,tau
```
kandrosov, line 1: maybe for the test it is better to have at least two dataset groups.

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,681 @@
...
+namespace {
+struct SourceDesc {
+    using Tau = tau_tuple::Tau;
+    using TauTuple = tau_tuple::TauTuple;
+    using SampleType = analysis::SampleType;
+
+    SourceDesc(const std::string& _name,const std::string&  _group_name, const std::vector<std::string>& _file_names,
+               double _weight, const std::set<std::string>& _disabled_branches) :
+        name(_name), group_name(_group_name), file_names(_file_names), disabled_branches(_disabled_branches), weight(_weight),
+        current_n_processed(0), files_n_total(_file_names.size()), entries_file(666), total_n_processed(0)
```
kandrosov, line 56: To make code parallelizable, an entry range should be passed. Since each file has a different size,  the range should be defined in relative values.

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,681 @@
...
+      (*current_tuple)().dataset_group_id = std::hash<std::string>{}(group_name);
+      return true;
+    }
+
+    const Tau& GetNextTau()
+    {
+      return current_tuple->data();
+    }
+
+    size_t GetNumberOfEvents() const { return entries_file; }
```
kandrosov, line 104: `GetNumberOfEvents` should return the total number of entries in all the files

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,681 @@
...
+          const std::string& file_name = file_names.at(*current_file_index);
+          current_tuple.reset();
+          current_file = root_ext::OpenRootFile(file_name);
+          current_tuple = std::make_shared<TauTuple>("taus", current_file.get(), true, disabled_branches);
+          entries_file = current_tuple->GetEntries();
+          if(!entries_file)
+            throw analysis::exception("Root file inside %1% is empty.") % name;
+      }
+      ++total_n_processed;
+      current_tuple->GetEntry(current_n_processed++);
```
kandrosov, line 93: tau types that are not specified in the description of the source, should be skipped. Dropping them later would bias the mix.

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,681 @@
...
+      (*current_tuple)().dataset_id = std::hash<std::string>{}(name);
+      (*current_tuple)().dataset_group_id = std::hash<std::string>{}(group_name);
```
kandrosov, line 95: - computing hash for each entry could be computationally costly. Can be done in the constructor
- std::hash does not guarantee reproducibility between different machines/compilers/environments. The table of hash<->name correspondence should be stored in an auxiliary TTree or in a csv/json file.

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,681 @@
...
+        using boost::filesystem::is_regular_file;
+        using boost::split;
+
+        name = item.name;
+        std::cout << name << std::endl;
+        const std::string dir_pattern_str = item.Get<std::string>("dir");
+        const std::string file_pattern_str = item.Get<std::string>("file");
+        weight = item.Has("weight") ? item.Get<double>("weight") : 1;
+        const std::string tau_types_str = item.Get<std::string>("types");
+        split(tau_types, tau_types_str, boost::is_any_of(","));
```
kandrosov, line 158: Better to use `analysis:: SplitValueList` or `analysis::SplitValueListT`, which is a commodity wrapper around `boost::split`
https://github.com/cms-tau-pog/TauMLTools/blob/master/Core/interface/TextIO.h#L112-L125
```c++
tau_types = SplitValueListT<TauType, std::set<TauType>>(tau_types_str, false, ",");
```

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,681 @@
...
+
+
+struct EntryDesc {
+    using TauType = analysis::TauType;
+    using SampleType = analysis::SampleType;
+
+    std::string name;
+    std::map< std::string, std::vector<std::string>> data_files;
+    std::map< std::string, std::string> spectrum_files;
+    std::vector< std::string > tau_types;
```
kandrosov, line 136: - better to define as `std::set<TauType> tau_types;`
- everywhere in the code, to represent tau types use `TauType` instead of `std::string`
https://github.com/cms-tau-pog/TauMLTools/blob/master/Analysis/interface/AnalysisTypes.h#L53

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,681 @@
...
+        const std::string dir_pattern_str = item.Get<std::string>("dir");
+        const std::string file_pattern_str = item.Get<std::string>("file");
+        weight = item.Has("weight") ? item.Get<double>("weight") : 1;
+        const std::string tau_types_str = item.Get<std::string>("types");
+        split(tau_types, tau_types_str, boost::is_any_of(","));
+
+        const std::vector<std::string> types ={"e","jet","tau","mu","all"};
+        for(auto el: tau_types)
+          if(std::find(types.begin(), types.end(), el) == types.end())
+            throw analysis::exception("Tau type '%1%' does not exist") % el;
```
kandrosov, line 163: considering the changes proposed above, these lines would not be needed.

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,681 @@
...
+        for(const auto& sample_dir_entry : make_iterator_range(directory_iterator(base_input_path))) {
+          if(!is_directory(sample_dir_entry) || !regex_match(sample_dir_entry.path().string(), dir_pattern)) continue;
+
+          std::cout << sample_dir_entry << " - dataset" << std::endl;
+          const std::string dir_name = sample_dir_entry.path().filename().string();
+          const path spectrum_file(base_spectrum_dir + "/" + dir_name + ".root");
+
+            if(!is_regular_file(spectrum_file)){ //make sure SpectrumHist is found
+              throw analysis::exception("No spectrum file are found for entry '%1%'") % sample_dir_entry;
+              continue;
```
kandrosov, line 180: this line will never be reached and can be removed.

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,681 @@
...
+                                            std::numeric_limits<size_t>::max()};
+    run::Argument<unsigned> n_threads{"n-threads", "number of threads", 1};
+    run::Argument<unsigned> seed{"seed", "random seed to initialize the generator used for sampling", 1234567};
+    run::Argument<std::string> disabled_branches{"disabled-branches",
+                                                 "list of branches to disabled in the input tuples", ""};
+    run::Argument<std::string> path_spectrum{"input_spec", "input path with spectrums for all the samples"};
+    run::Argument<std::string> tau_ratio{"tau_ratio", "ratio of tau types in the final spectrum"};
+};
+
+namespace {
```
kandrosov, line 47: all classes can be defined inside namespace `analysis`

kandrosov (changes requested):
Partial review. To be continued.

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,773 @@
...
+        current_tau_type = analysis::GenMatchToTauType(gen_match, sample_type);
+        ++total_n_input;
+      }
+      while (tau_types.find(current_tau_type) == tau_types.end());
+      ++total_n_processed;
+      (*current_tuple)().tauType = static_cast<Int_t>(current_tau_type);
+      (*current_tuple)().dataset_id = dataset_hash;
+      (*current_tuple)().dataset_group_id = datagroup_hash;
+      return true;
+    }
```
shedprog, line 111: @kandrosov I found out that with this approach at some point of iteration virtual memory usage exceeds 400Gb and this jobs is killed by task manager. Is this some bug from my side or opening all datasets in parallel can potentially exceed this limit?

shedprog: The RAM exceeds 14 GB...

kandrosov: How many dataset groups do you have? Having about 10-20 groups (i.e. 10-20 simultaneously opened files) should not require that much memory.

shedprog: I have 7 data groups, but the concept of the code was to have all data sets opened in parallel, that is ~87 datasets

kandrosov: hm... no, this concept should be changed. Only one file from each dataset group should be opened at the time...

shedprog: I see.... in this case everything should be re-done conceptually, I thought in the beginning we agreed about such concept

shedprog: In case of "one file from datagroup opened" how do we mix them within a single data group they just will be sequentially ordered.

kandrosov: Hm... in the old implementation of ShuffleMerge only one file from one bin of the given dataset group was opened at the time. My apologies, I probably have missed the point that you wanted to change that concept... 
I remember we have discussed that one of the main limitations of having binned dataset, is the necessity of having opened multiple files per dataset group. Switching to the spectral shuffle&merge was a solution for this issue (among the other issues as well) - with it, dataset group is read consecutively (as in the past) supposing that all datasets in it having equivalent properties, but this time only 1 file per dataset group must be opened at the time.

kandrosov: > In case of "one file from datagroup opened" how do we mix them within a single data group they just will be sequentially ordered.

If all datasets must be mixed, then "dataset group" abstraction does not make much sense - the only benefit I see from combining datasets into groups is to reduce the number of sources that must be mixed...

shedprog: > > In case of "one file from datagroup opened" how do we mix them within a single data group they just will be sequentially ordered.
> 
> If all datasets must be mixed, then "dataset group" abstraction does not make much sense - the only benefit I see from combining datasets into groups is to reduce the number of sources that must be mixed...

yes, my bad.... I also had so many doubts about the concept of data group and decided in the end that it was just a logical separation, can you please confirm that now my understanding of the concept is right:
1. We are randomly the order of the files within one data group.
2. We have SourceDesc per every data group, and we are "walking" within shuffled files inside it sequentially.
3. We ensure homogenous occurrence of data group and hope that it also will be relatively homogenous over datasets.
4. Now, it also becomes more clear that this parallelization that we implemented with start and end ratio within a single file actually add additional randomization for dataset occurrence, so, it should be effective not only for saving time....

kandrosov: Yes, I confirm that it is also my current understanding. Small remark on point 3. If we group samples with the same physics signature, the requirement of homogeneity of samples from this group across the output dataset is not strictly necessary. Therefore, if we are careful in the definition of the groups, we don't need to "hope that it also will be relatively homogenous over datasets".

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,773 @@
...
+            else
+                ++(*current_file_index);
+            if(*current_file_index >= file_names.size())
+                throw analysis::exception("The expected number of events = %1% is bigger than the actual number of"
+                                          " end point events in source '%2%'.") % entries_file % name;
+            const std::string& file_name = file_names.at(*current_file_index);
+            dataset_hash = dataset_hash_arr.at(*current_file_index);
+            current_tuple.reset();
+            current_file = root_ext::OpenRootFile(file_name);
+            current_tuple = std::make_shared<TauTuple>("taus", current_file.get(), true, disabled_branches);
```
shedprog, line 96: @kandrosov small question: does it grantee that memory allocated for current_file and current_tuple is cleaned properly after each iteration?

The problem with huge virtual memory consumption was solved by switching to the "SourceDesc per data group" concept, but I still do not understand why RES still incises consumption after each iteration, it is not much, but it linearly increases. If I just move this 3 lines to the constructor of SourceDesc (open for one file always) - that is fine (it is stable).

kandrosov: If the smart pointer is reset, the corresponding destructors must be called. In [SmartTree destructor](https://github.com/cms-tau-pog/TauMLTools/blob/master/Core/interface/SmartTree.h#L222-L223) the TTree is deleted. So if there are some memory leaks, it should be in the root code (as far as I can see)... You might try to add these two lines before opening a new file:
```c++
if(current_file)
    current_file->Close();
```
Beyond that, I don't know what else can we do reduce the leaks...

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,775 @@
...
+      Generator gen(args.seed());
+
+      for(const auto& e : entries) {
+        const std::string& file_name = e.first;
+        const std::vector<EntryDesc>& entry_list = e.second;
+        std::cout << "Processing:";
+        for(const auto& entry : entry_list)
+          std::cout << ' ' << entry.name;
+        std::cout << "\nOutput: " << file_name << std::endl;
+        auto output_file = root_ext::CreateRootFile(file_name, ROOT::kLZ4, 4);
```
kandrosov, line 677: Better to have compression level and algorithm as an argument. I'm thinking if it makes sense to switch to the max compression (i.e. LZMA with lvl=9) to minimize the size of the shuffled dataset.

Analysis/config/2018/training_inputs_MC.cfg
```diff
@@ -0,0 +1,7 @@
+DY: dir=(DY.*) file=eventTuple.* types=jet,e,mu,tau
+TT: dir=(TT.*) file=eventTuple.* types=jet,e,mu,tau
+W:  dir=(W1Jets.*|W2Jets.*|W3Jets.*|W4Jets.*|WJets.*) file=eventTuple.* types=jet,e,mu,tau
+QCD: dir=(QCD.*) file=eventTuple.* types=jet
+Higgs: dir=(ttHTo.*|GluGluHTo.*|ZHTo.*|WplusHTo.*|WminusHTo.*|VBF.*|GluGlu.*) file=eventTuple.*  types=jet,e,mu,tau
```
kandrosov, line 5: - from Higgs samples, I suggest taking only tau
- Add embedded samples (taking emb_tau and emb_jet)

kandrosov (changes requested):

Analysis/interface/AnalysisTypes.h
```diff
@@ -71,11 +72,15 @@ inline TauType GenMatchToTauType(GenLeptonMatch gen_match, SampleType sample_typ
...
-        if(sample_type == SampleType::Embedded) return TauType::emb;
+        if(sample_type == SampleType::Embedded) return TauType::emb_tau;
     } else if(gen_match == GenLeptonMatch::NoMatch) {
         if(sample_type == SampleType::MC) return TauType::jet;
         if(sample_type == SampleType::Data) return TauType::data;
-    }
+        if(sample_type == SampleType::Embedded) return TauType::emb_jet;
+    } else if(sample_type == SampleType::Embedded && (gen_match == GenLeptonMatch::Muon
+              || gen_match == GenLeptonMatch::TauMuon)) return TauType::emb_mu;
+      else if(sample_type == SampleType::Embedded && gen_match == GenLeptonMatch::TauElectron)
```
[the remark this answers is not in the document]
shedprog: I have written in this way because there were no gen_match == GenLeptonMatch::Electron within embedded sample

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,775 @@
...
+};
+
+struct SourceDesc {
+    using Tau = tau_tuple::Tau;
+    using TauType = analysis::TauType;
+    using TauTuple = tau_tuple::TauTuple;
+    using SampleType = analysis::SampleType;
+    using Generator = std::mt19937_64;
+
+    SourceDesc(const std::string& name_, const Int_t&  _group_hash,
```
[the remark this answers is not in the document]
shedprog: The practical reason was that inside our TauTuple  `*id` was allocated as `Int_t` i.e: `VAR(Int_t, dataset_id)`

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -0,0 +1,775 @@
...
+    return tartget_hist;
+  }
+
+  static std::pair<std::vector<int>,std::vector<int>> GetBinMap(const std::shared_ptr<TH2D>& newbins,
+                                                                const std::shared_ptr<TH2D>& oldbins)
+  {
+    // check over x axis
+    std::vector<int> x_map;
+    for(Int_t new_x=0; new_x<=newbins->GetNbinsX(); new_x++){
+      float up_x = (float)newbins->GetXaxis()->GetBinUpEdge(new_x);
```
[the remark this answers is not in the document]
shedprog: > "why conversion to float is necessary?"
Sometimes `->GetBinUpEdge` method when for instance having edge at 1.5 it returns 1.5(00).....1 when using Double_t. Conversion to float was a dirty step to avoid it.

Analysis/bin/ShuffleMergeSpectral.cxx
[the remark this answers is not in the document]
kandrosov: Ok, I see. I think you can change type to `ULong64_t` and disable branch when opening original files (anyhow dataset_id is defined only after ShuffleMerge).

Analysis/interface/AnalysisTypes.h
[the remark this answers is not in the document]
kandrosov: Have you see any `gen_match == GenLeptonMatch::Muon` within embedded sample? In theory, there also shouldn't be any of them. Ele and mu sections should be consistent. I agree that removing `gen_match == GenLeptonMatch::Muon` should be a better choice.

Core/src/RootExt.cpp
```diff
@@ -84,6 +86,52 @@ ClassInheritance FindClassInheritance(const std::string& class_name)
...
+
+    static const auto get_new_bin = [](const TAxis* old_axis, const TAxis* new_axis, int bin_id_old) {
+        const double old_low_edge = old_axis->GetBinLowEdge(bin_id_old);
+        const double old_up_edge = old_axis->GetBinUpEdge(bin_id_old);
+        const int bin_low_new = new_axis->FindFixBin(old_low_edge);
+        const int bin_up_new = new_axis->FindFixBin(old_up_edge);
+
+        if(bin_low_new != bin_up_new
+          && !(fabs(old_up_edge-new_axis->GetBinUpEdge(bin_low_new)) <= std::numeric_limits<double>::epsilon()*2))
+          throw analysis::exception("Uncompatible bin edges");
```
shedprog, line 108: Few comments about this line:
1) e.g: lets say we have the following bins: 0.0, 1.0, 2.0... in the situation when we have old_low_edge= 0.8 and old_up_edge=1.0, `new_axis->FindFixBin(1.0)` will return 1, but `new_axis->FindFixBin(1.0)` will return 2, I do not know why but this method include edge in the next bin.... 
2) in this case we need to check if upper limit of the bin 1 corresponds to old_up_edge... but that does not work because `GetBinUpEdge(1)` returns it with numeric type inaccuracy 1.00000000000000009..... `old_axis->GetBinUpEdge(bin_id_old)` also was returned with numerical inaccuracy, so the difference should be within `numeric_limits<double>::epsilon() * 2` (it is interesting that difference is alwaysexactly two epsolons)

kandrosov: - If bins are defined explicitly (as an array) there is no numeric inaccuracy involved because there were no floating-point operations applied to the number. I don't understand why for input bins = { 0., 1., 2. }, GetBinUpEdge(1) would return something different from 1.
- Indeed, all bin intervals in the TH histograms are in the form [a, b), i.e. lower bound is included, but the upper bound is not. I forgot to account for it in the draft. This additional condition should solve that. However, epsilon is a relative precision, so I propose to check the compatibility in the following way:
```c++
const double new_up_edge = old_axis->GetBinUpEdge(bin_low_new);
if(bin_low_new != bin_up_new
        && !(std::abs(old_up_edge-new_up_edge) <= std::numeric_limits<double>::epsilon() * std::abs(old_up_edge+new_up_edge) * 2))
    throw analysis::exception("Uncompatible bin edges");
```

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -16,6 +16,8 @@
 #include "TauMLTools/Core/interface/PropertyConfigReader.h"
 #include "TauMLTools/Core/interface/ProgressReporter.h"
 #include "TauMLTools/Analysis/interface/TauTuple.h"
+#include <limits>
+typedef std::numeric_limits< double > dbl;
```
kandrosov, line 20: `typedef std::numeric_limits< double > dbl;` -> `using dbl = std::numeric_limits<double>;`

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -46,30 +48,30 @@ struct Arguments {
     run::Argument<double> start_entry{"start-entry", "starting ratio from which file will be processed", 0};
     run::Argument<double> end_entry{"end-entry", "end ratio until which file will be processed", 1};
     run::Argument<double> exp_disbalance{"exp-disbalance", "maximal expected disbalance between low pt and high pt regions",0};
+    run::Argument<unsigned> compression_level{"compression-level", "compression level of output file", 4};
```
kandrosov, line 51: For completeness, the compression algorithm should also be passed as an argument.

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -688,8 +650,9 @@ class ShuffleMergeSpectral {
           const auto& tau = processor.GetNextTau();
           n_processed++;
           (*output_tuple)() = tau;
-          output_tuple->Fill();
-          if(n_processed % 1000 == 0){
+          if(parity==3) output_tuple->Fill();
+      	  else if(tau.evt % 2 == parity) output_tuple->Fill();
```
kandrosov, line 654: `if(parity == 3 || (tau.evt % 2 == parity)) output_tuple->Fill();`

Core/interface/RootExt.h
```diff
@@ -12,6 +12,8 @@ This file is part of https://github.com/hh-italian-group/TauMLTools. */
 #include <TMatrixD.h>
 #include <TFile.h>
 #include <Compression.h>
+#include <TAxis.h>
```
kandrosov, line 15: Including TAxis is not necessary in RootExt.h. If needed, include it in RootExt.cpp.

Core/src/RootExt.cpp
```diff
@@ -12,6 +12,8 @@ This file is part of https://github.com/hh-italian-group/TauMLTools. */
 #include <TMatrixD.h>
 #include <TFile.h>
 #include <Compression.h>
+#include <TAxis.h>
+#include <TH2.h>
```
kandrosov, line 16: TH2.h is already included in RootExt.h.

kandrosov (changes requested):

Core/src/RootExt.cpp
```diff
@@ -103,9 +102,10 @@ void RebinAndFill(TH2& new_hist, const TH2& old_hist)
         const int bin_low_new = new_axis->FindFixBin(old_low_edge);
         const int bin_up_new = new_axis->FindFixBin(old_up_edge);
 
+        const double new_up_edge = new_axis->GetBinUpEdge(bin_low_new);
```
shedprog, line 105: I guess here  `new_axis->GetBinUpEdge(bin_low_new)` is needed not `old_axis->GetBinUpEdge(bin_low_new)`

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -48,6 +46,7 @@ struct Arguments {
     run::Argument<double> start_entry{"start-entry", "starting ratio from which file will be processed", 0};
     run::Argument<double> end_entry{"end-entry", "end ratio until which file will be processed", 1};
     run::Argument<double> exp_disbalance{"exp-disbalance", "maximal expected disbalance between low pt and high pt regions",0};
+    run::Argument<std::string> compression_algo{"compression-algo","ZLIB, LZMA, LZ4","LZ4"};
```
kandrosov, line 49: put as default LZMA with lvl=9. We will use the first big shuffle&merge production to test if decompression is a bottleneck for the training. If yes, we can consider another compression, if no - better to have smaller files.

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -724,6 +718,12 @@ class ShuffleMergeSpectral {
         std::cout << ".\n";
     }
 
+    static ROOT::ECompressionAlgorithm compAlg(const std::string& comp_string) {
+      if("ZLIB") return ROOT::kZLIB;
+      if("LZMA") return ROOT::kLZMA;
+      if("LZ4") return ROOT::kLZ4;
```
kandrosov, line 724: The current implementation will result in undefined behaviour if `comp_string` is other than the considered 3 options.

kandrosov (changes requested):

PR #31, opened 2020-10-26
adding validation tool (KS test)

lguzzi:
PR to add the vaidation tool. The script loads a list of files into a RDataFrame, splits the dataframe into different chunks and then runs a Kolmogorov-Smirnov test between the first chunk and the rest to check the compatibility of some distributions.

The distributions considered for the KS test are:
- lepton_gen_match, and:
  - tau_pt and tau_eta for each gen match value
- sample_type distribution, and:
  - tau_pt and tau_eta for each sample type value
- dataset_group_id distribution, and:
  - tau_pt, tau_eta for each group id value
  - dataset_id for each group id value

The script creates a root file storing the chunk histograms, a json file with the p-values and the pdf files of the KS comparison. 
You can find and example at /afs/cern.ch/user/l/lguzzi/public/TauPOG/validation_tool_test (run on /eos/cms/store/group/phys_tau/TauML/prod_2018_v1/full_tuples/WJetsToLNu_HT-2500ToInf/eventTuple_1-1.root)

### Time tests
Here are some timing test I run on an lxplus machine:
- root-version: 6.20/02
- dataset: WJetsToLNu_HT-1200To2500
  - size: 42 GB
  - events: 15890103 (15 milion)

```
source /cvmfs/sft.cern.ch/lcg/views/LCG_97apython3/x86_64-centos7-clang10-opt/setup.sh
python -m cProfile validation_tool.py --input "/eos/cms/store/group/phys_tau/TauML/prod_2018_v1/full_tuples/WJetsToLNu_HT-1200To2500/*.root" --output test-v5 --legend
```
Run time after a fresh login (subsequent calls are about twice as fast):
1 thread: 106 s
4 thread: 47 s
10 thread:  19 s

The most time consuming function is load_histograms, which takes about 7s per call running on one thread, 2s per call  running on four threads and 1s per call on ten threads (13 calls in total).

### DIFF
--- a/Production/scripts/validation_tool.py
+++ b/Production/scripts/validation_tool.py
@@ -132,7 +132,7 @@ if __name__ == '__main__':
 
   ## first, run on plain columns
   dataframe = ROOT.RDataFrame('taus', input_files)
-  run_validation(dataframe = dataframe, pwd = main_dir, branches = ['lepton_gen_match', 'sampleType', 'dataset_group_id'])
+  run_validation(dataframe = dataframe, pwd = main_dir, branches = ['lepton_gen_match', 'sampleType', 'dataset_group_id', 'dataset_id'])
 
   ## then, group by tau type
   tau_type_dataframes = groupby(dataframe = dataframe, by = 'lepton_gen_match')
@@ -148,6 +148,11 @@ if __name__ == '__main__':
   group_id_dataframes = groupby(dataframe = dataframe, by = 'dataset_group_id')
   for ii, df in group_id_dataframes.iteritems():
     run_validation(dataframe = df, pwd = '/'.join([main_dir, 'dataset_group_id', str(ii)]), branches = ['tau_pt', 'tau_eta', 'dataset_id'])
+
+  ## then, group by dataset id
+  group_id_dataframes = groupby(dataframe = dataframe, by = 'dataset_id')
+  for ii, df in group_id_dataframes.iteritems():
+    run_validation(dataframe = df, pwd = '/'.join([main_dir, 'dataset_id', str(ii)]), branches = ['tau_pt', 'tau_eta'])
 
   OUTPUT_ROOT.Close()
   json.dump(JSON_DICT, OUTPUT_JSON, indent = 4)
--- a/Production/scripts/validation_tool.py
+++ b/Production/scripts/validation_tool.py
@@ -12,7 +12,7 @@ import glob
 import json
 from collections import OrderedDict
 
-parser.add_argument('--input'       , required = True, type = str, help = 'input file. Accepts glob patterns')
+parser.add_argument('--input'       , required = True, type = str, help = 'input file. Accepts glob patterns (use quotes)')
 parser.add_argument('--output'      , required = True, type = str, help = 'output file name')
 parser.add_argument('--pdf'         , default  = None, type = str, help = 'output pdf directory')
 parser.add_argument('--json'        , required = True, type = str, help = 'output json file name')
--- a/Production/scripts/validation_tool.py
+++ b/Production/scripts/validation_tool.py
@@ -55,7 +55,7 @@ def groupby(dataframe, by):
   return {tt: dataframe.Filter('{} == {}'.format(by, tt)) for tt in types}
 
 def get_histos(dataframe, branch, norm = False):
-  size = int(dataframe.Count())
+  size = dataframe.Count().GetValue()
   sub_size = 1 + size // N_SPLITS
   subframes = [dataframe.Range(ii*sub_size, (ii+1)*sub_size) for ii in range(N_SPLITS)]
   
--- a/Production/scripts/validation_tool.py
+++ b/Production/scripts/validation_tool.py
@@ -51,6 +51,7 @@ def groupby(dataframe, by):
   hist = _.GetValue()
   hist.ClearUnderflowAndOverflow()
   types = list(set([round(hist.GetBinCenter(jj)) for jj in range(hist.GetNbinsX()) if hist.GetBinContent(jj)]))
+  types = [int(tt) for tt in types]
 
   return {tt: dataframe.Filter('{} == {}'.format(by, tt)) for tt in types}
 
--- a/Production/scripts/validation_tool.py
+++ b/Production/scripts/validation_tool.py
@@ -64,8 +64,9 @@ def get_histos(dataframe, branch, norm = False):
   hptrs = [sf.Histo1D(model, branch) for sf in subframes]
   histos = [hh.GetValue().Clone() for hh in hptrs]
   
-  for hh in histos:
+  for ii, hh in enumerate(histos):
     hh.SetTitle(branch)
+    hh.SetName('_'.join([hh.GetName(), ii]))
     hh.Sumw2()
     hh.ClearUnderflowAndOverflow()
     if norm and hh.Integral():
--- a/Production/scripts/validation_tool.py
+++ b/Production/scripts/validation_tool.py
@@ -63,10 +63,10 @@ def get_histos(dataframe, branch, norm = False):
   model = (branch, "") + BINS[branch]
   hptrs = [sf.Histo1D(model, branch) for sf in subframes]
   histos = [hh.GetValue().Clone() for hh in hptrs]
-  
+
   for ii, hh in enumerate(histos):
     hh.SetTitle(branch)
-    hh.SetName('_'.join([hh.GetName(), ii]))
+    hh.SetName('_'.join([hh.GetName(), str(ii)]))
     hh.Sumw2()
     hh.ClearUnderflowAndOverflow()
     if norm and hh.Integral():
--- a/Production/scripts/validation_tool.py
+++ b/Production/scripts/validation_tool.py
@@ -1,6 +1,4 @@
 from __future__ import print_function
-import sys; PYTHON_MAJOR = int(sys.version_info.major)
-dict_iterator = 'items' if PYTHON_MAJOR == 3 else 'iteritems'
 
 import ROOT
 import glob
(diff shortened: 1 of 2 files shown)


Production/scripts/validation_tool.py
```diff
@@ -0,0 +1,159 @@
...
+    
+    save_histos(histos, fdir = '/'.join([pwd, branch]), pvalues = pvalues)
+
+if __name__ == '__main__':
+  print '[INFO] reading files', args.input
+  input_files = ROOT.std.vector('std::string')()
+  for file in glob.glob(args.input):
+    input_files.push_back(str(file))
+
+  main_dir = 'KS_test'
```
kandrosov, line 131: I suggest to use `args.output` as the main dir and store output root file inside it.

Production/scripts/validation_tool.py
```diff
@@ -0,0 +1,159 @@
...
+  hist = _.GetValue()
+  hist.ClearUnderflowAndOverflow()
+  types = list(set([round(hist.GetBinCenter(jj)) for jj in range(hist.GetNbinsX()) if hist.GetBinContent(jj)]))
+  types = [int(tt) for tt in types]
+
+  return {tt: dataframe.Filter('{} == {}'.format(by, tt)) for tt in types}
+
+def get_histos(dataframe, branch, norm = False):
+  size = int(dataframe.Count())
+  sub_size = 1 + size / N_SPLITS
```
kandrosov, line 60: `sub_size = 1 + size // N_SPLITS` to be python3-friendly

Production/scripts/validation_tool.py
```diff
@@ -0,0 +1,159 @@
...
+
+parser.add_argument('--input'       , required = True, type = str, help = 'input file. Accepts glob patterns')
+parser.add_argument('--output'      , required = True, type = str, help = 'output file name')
+parser.add_argument('--pdf'         , default  = None, type = str, help = 'output pdf directory')
+parser.add_argument('--json'        , required = True, type = str, help = 'output json file name')
+parser.add_argument('--nsplit'      , default  = 5   , type = str, help = 'number of chunks per file')
+parser.add_argument('--pvthreshold' , default  = .05 , type = str, help = 'threshold of KS test (above = ok)')
+
+parser.add_argument('--visual', action = 'store_true', help = 'Won\'t run the script in batch mode')
+parser.add_argument('--legend', action = 'store_true', help = 'Draw a TLegent on canvases')
```
kandrosov, line 23: add argument to specify number of threads and then set implicit multi-threading for ROOT:
```python
if args.n_threads > 1:
    ROOT.ROOT.EnableImplicitMT(args.n_threads)
```

Production/scripts/validation_tool.py
```diff
@@ -0,0 +1,159 @@
...
+
+  ## then, group by dataset group id
+  group_id_dataframes = groupby(dataframe = dataframe, by = 'dataset_group_id')
+  for ii, df in group_id_dataframes.iteritems():
+    run_validation(dataframe = df, pwd = '/'.join([main_dir, 'dataset_group_id', str(ii)]), branches = ['tau_pt', 'tau_eta', 'dataset_id'])
+
+  ## then, group by dataset id
+  group_id_dataframes = groupby(dataframe = dataframe, by = 'dataset_id')
+  for ii, df in group_id_dataframes.iteritems():
+    run_validation(dataframe = df, pwd = '/'.join([main_dir, 'dataset_id', str(ii)]), branches = ['tau_pt', 'tau_eta'])
```
kandrosov, line 155: it will be faster if `run_validation` is split into two parts:
- first, get histogram pointers for all data frames and all branches
- then do another loop over data frames and branches, and transform pointers into histograms by calling GetValue.

In this way, it would require only one passage over the input files. For a large dataset, multiple passages would considerably slow down the processing time.

kandrosov (changes requested):

Production/scripts/validation_tool.py
lguzzi: Unfortunatelly, "Range was called with ImplicitMT enabled. Multi-thread ranges are not supported."
Also, see https://root-forum.cern.ch/t/rdataframe-range-does-not-work-with-multi-thread/40466/2. I don't know if still holds for our root release
I don't know if there is any workaround

lguzzi: full traceback:
```
Traceback (most recent call last):
  File "/cvmfs/cms.cern.ch/slc7_amd64_gcc700/external/python/2.7.14-pafccj/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/cvmfs/cms.cern.ch/slc7_amd64_gcc700/external/python/2.7.14-pafccj/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/cvmfs/cms.cern.ch/slc7_amd64_gcc700/external/python/2.7.14-pafccj/lib/python2.7/profile.py", line 610, in <module>
    main()
  File "/cvmfs/cms.cern.ch/slc7_amd64_gcc700/external/python/2.7.14-pafccj/lib/python2.7/profile.py", line 603, in main
    runctx(code, globs, None, options.outfile, options.sort)
  File "/cvmfs/cms.cern.ch/slc7_amd64_gcc700/external/python/2.7.14-pafccj/lib/python2.7/profile.py", line 77, in runctx
    prof = prof.runctx(statement, globals, locals)
  File "/cvmfs/cms.cern.ch/slc7_amd64_gcc700/external/python/2.7.14-pafccj/lib/python2.7/profile.py", line 444, in runctx
    exec cmd in globals, locals
  File "validation_tool.py", line 136, in <module>
    run_validation(dataframe = dataframe, pwd = main_dir, branches = ['lepton_gen_match', 'sampleType', 'dataset_group_id', 'dataset_id'])
  File "validation_tool.py", line 111, in run_validation
    histos = get_histos(dataframe, branch = branch, norm = True)
  File "validation_tool.py", line 61, in get_histos
    subframes = [dataframe.Range(ii*sub_size, (ii+1)*sub_size) for ii in range(N_SPLITS)]
TypeError: none of the 2 overloaded methods succeeded. Full details:
  ROOT::RDF::RInterface<ROOT::Detail::RDF::RRange<ROOT::Detail::RDF::RLoopManager>,void> ROOT::RDF::RInterface<ROOT::Detail::RDF::RLoopManager,void>::Range(unsigned int begin, unsigned int end, unsigned int stride = 1) =>
    Range was called with ImplicitMT enabled. Multi-thread ranges are not supported. (C++ exception of type runtime_error)
  ROOT::RDF::RInterface<ROOT::Detail::RDF::RRange<ROOT::Detail::RDF::RLoopManager>,void> ROOT::RDF::RInterface<ROOT::Detail::RDF::RLoopManager,void>::Range(unsigned int end) =>
    takes at most 1 arguments (2 given)
```

kandrosov: I see. We planned that starting from the next steps (after the ShuffleMerge) we'll use LCG environment instead of the CMSSW-one. Probably it can be already the case for this tool - like this it could be run within the same environment as all further scripts.
Please, try to source LCG environment (as suggested in README.md) and try to run the script there. 
```sh
source /cvmfs/sft.cern.ch/lcg/views/LCG_97apython3/x86_64-centos7-clang10-opt/setup.sh
```

Production/scripts/validation_tool.py
kandrosov: hm.. actually I misread the code, `main_dir` is used for the path inside the root file. Ignore the previous comment.

Production/scripts/validation_tool.py
lguzzi: looks like it is still unsupported

kandrosov: ok... no implicit multi-threading then...

Production/scripts/validation_tool.py
lguzzi: I think I can work something out using filters instead of ranges,

kandrosov: 26 times seems a bit too much. I agree that we can implement things without ranges then (and, as a bonus, we'll be able to run with implicit multithreading). Anyhow, we need to process the entire dataset in order to do validation, so probably ranges (in their current implementation) is not something we want.
I don't think you need filters for that:
- define variable chunk_id: `df = df.Define("chunk_id", "rdfentry_ % {}".format(n_chunks))`
- create 2D histograms instead 1D, where the second dimension is filled with chunk_id (for pt/eta it will be 3D histogram then)
- After 2D histograms are created and filled, convert each slice into 1D and the rest is equivalent to how it is implemented currently in v2.

kandrosov: Sorry, small correction for chunk_id definition:
```python
df = df.Define("chunk_id", "rdfentry_ * {} / {}".format(n_chunks, total_n_entries))
```

lguzzi: I've updated the [validation tool](https://github.com/lguzzi/TauMLTools/blob/dev/Production/scripts/validation_tool-v5.py) in my dev branch. Before merging to master, could you please explain why shouldn't I use 
```
df = df.Define("chunk_id", "rdfentry_ % {}".format(n_chunks))
```

kandrosov: With `%` your chunks will be distributed through the whole dataset. E.g. with two chunks you'll be testing compatibility between even and odd entries. It is not what we want. We want to check compatibility between different parts of the dataset. E.g. with two chunks we want to test that the first 50% of taus having the same properties as the last 50% of taus.
Typical batch size is about 100-200 taus, so even if by some mistake taus with even entry id would have different properties wrt to the taus with odd entry id - it will be transparent for the training. On the other hand, if the first half of the dataset has different properties, compared to the second part of the dataset - it is a problem: the likelihood minimums will be different in different parts => minimization will not be able to converge.

lguzzi: Thank you, I've updated the script with your suggestions

kandrosov (commented):
Thank you @lguzzi! The current version code looks good. To close this PR, could you please:
- Post here timing to run the tool on a relatively large dataset (at least O(10 GB)) for the future references.
- Add instructions on how to run the validation into README.md

kandrosov (approved):

PR #32, opened 2020-11-10
update validation tool

lguzzi:
Updating validation tool:

- now the input is a directory and all '*.root' files inside that directory and all of its subdirectories are considered (using os.walk module)
- the lepton_gen_match variable has been changed to tauType
- now can handle hashed dataset group and dataset ids (two additional arguments added: path to the dataset group and dataset id json files)
-updated readme

tested running:
```
 python TauMLTools/Production/scripts/validation_tool.py  --input /afs/cern.ch/work/m/myshched/public/shuffle/ --output test-dir_2 --n_threads 10  --legend --id_json /afs/cern.ch/work/m/myshched/public/shuffle/dataset_hash.json --group_id_json /afs/cern.ch/work/m/myshched/public/shuffle/datagroup_hash.json --nsplit 5
```

### DIFF
--- a/Production/scripts/validation_tool.py
+++ b/Production/scripts/validation_tool.py
@@ -42,7 +42,7 @@ PVAL_THRESHOLD = args.pvthreshold
 BINS = {
   'tau_pt'    : (50, 0, 5000),
   'tau_eta'   : (5, -3.2, 3.2),
-  'tauType' : (4, 0, 4),
+  'tauType' : (10, 0, 10),
   'sampleType': (20, -1, 19),
 }
 
--- a/Production/scripts/validation_tool.py
+++ b/Production/scripts/validation_tool.py
@@ -1,9 +1,9 @@
 from __future__ import print_function
 
 import ROOT
-import glob
 import json
 from collections import OrderedDict
+import os
 
 import argparse
 parser = argparse.ArgumentParser('''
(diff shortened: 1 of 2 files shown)


Production/scripts/validation_tool.py
```diff
@@ -116,40 +115,60 @@ def to_2D(histo, vbin):
...
     ROOT.ROOT.EnableImplicitMT(args.n_threads)
 
   input_files = ROOT.std.vector('std::string')()
   
-  for file in glob.glob(args.input):
-    input_files.push_back(str(file))
+  for root, dirs, files in os.walk(args.input):
+    for ff in files:
+      if not '.root' in ff: continue
+      input_files.push_back('/'.join([root, ff]))
```
kandrosov, line 128: - I suggest first to fill a list, then sort it and only after that convert to std::vector. In this way we don't need to rely on the order on which os.walk returns the files. 
- Add regex filter as an argument (with default value `.*\.root$`)

kandrosov (changes requested):

Production/scripts/validation_tool.py
lguzzi: done ed0e96117c3bac96bef3ea9ccc69e838e6f3a683

Production/scripts/validation_tool.py
```diff
@@ -37,14 +38,12 @@
...
-## binning of tested variables
+## binning of tested variables (dataset group id and dataset id are guessed from jsons)
 BINS = {
   'tau_pt'    : (50, 0, 5000),
   'tau_eta'   : (5, -3.2, 3.2),
-  'lepton_gen_match' : (20, -1, 19),
-  'sampleType': (20, -1, 19),      
-  'dataset_id': (20, -1, 19),
-  'dataset_group_id': (20, -1, 19),
+  'tauType' : (4, 0, 4),
```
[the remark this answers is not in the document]
lguzzi: fixed ed0e96117c3bac96bef3ea9ccc69e838e6f3a683

kandrosov (approved):

lguzzi:
I have updated the script and added an additional argument (--use_dataset_id) to make the KS test on dataset_id optional

PR #33, opened 2020-11-20
Law tool (ShuffleMergeSpectral)

lguzzi:
Tool to run ShuffleMergeSpectral jobs on condor. The tool is a very simple implementation of a [law](https://github.com/riga/law) workflow.

Each job is a luigi task creating a pipe (python subprocess module) with the ShuffleMergeSpectral command properly set.
Jobs are defined using the --n-jobs, --start-entry and --end-entry arguments, that is, jobs will range from --start-entry to --end-entry with a step equal to (end-entry - start-entry) / n-jobs.

The script is still missing the validation part, which will be added as a task which will follow the shuffle and merge tasks.  

NOTE: I haven't tested the script on real jobs, yet (currently running). I am creating this PR to get a feedback and possible suggestions.

tested on lxplus on top of f866a4688df26764f7013a17784364929c5838dd running (*)
```sh
law run ShuffleMergeSpectral --version v1 --cfg Analysis/config/2018/training_inputs_MC.cfg --input-path /eos/cms/store/group/phys_tau/TauML/prod_2018_v1/full_tuples --output-path test.root --mode MergeAll --n-threads 10 --disabled-branches "trainingWeight, tauType" --input-spec /afs/cern.ch/work/m/myshched/public/root-tuple-v2-hists --pt-bins "20, 30, 40, 50, 60, 70, 80, 90, 100, 1000" --eta-bins "0., 0.6, 1.2, 1.8, 2.4" --tau-ratio "jet:1, e:1, mu:1, tau:1" --exp-disbalance 100 --start-entry 0.0 --end-entry 0.001 --n-jobs 10
```

Please note that the arguments of the law command are the same of the ShuffleMergeSpectral command (plus --n-jobs), but --input and --output have been renamed to --input-path and --output-path.

(*) temporary change *TauType::emb* to *TauType::emb_tau* in https://github.com/cms-tau-pog/TauMLTools/blob/master/Analysis/bin/TrainingTupleProducer.cxx#L306


Production/scripts/ShuffleMergeLaw/analysis/framework.py
```diff
@@ -0,0 +1,90 @@
+# coding: utf-8
```
kandrosov, line 1: - Move the whole project into Analysis/law/ShuffleMerge
- `framework.py` seems to be independent on ShuffleMerge - put it directly in `Analysis/law`, so it can be reused for other job submissions in the future.
- is it necessary to have an empty `__init__.py`?

Production/scripts/ShuffleMergeLaw/analysis/tasks.py
```diff
@@ -0,0 +1,64 @@
...
+  input_spec        = luigi.Parameter(description = '')
+  tau_ratio         = luigi.Parameter(description = '')
+  disabled_branches = luigi.Parameter(description = 'disabled-branches list of branches to disabled in the input tuples',
+                                      default = '')
+  n_threads         = luigi.IntParameter(description = 'number of threads', default = 1)
+  exp_disbalance    = luigi.IntParameter(description = '')
+  n_jobs            = luigi.IntParameter(description = 'number of HTCondor jobs to run')
+  start_entry       = luigi.FloatParameter(description = 'starting point')
+  end_entry         = luigi.FloatParameter(description = 'ending point')
+
```
kandrosov, line 29: there will be also a parity-related argument. Please, sync modifications with Mykyta.

Production/scripts/ShuffleMergeLaw/analysis/tasks.py
```diff
@@ -0,0 +1,64 @@
...
+  n_jobs            = luigi.IntParameter(description = 'number of HTCondor jobs to run')
+  start_entry       = luigi.FloatParameter(description = 'starting point')
+  end_entry         = luigi.FloatParameter(description = 'ending point')
+  
+  def create_branch_map(self):
+    step = 1. * (self.end_entry - self.start_entry) / self.n_jobs
+    return {i: (round(self.start_entry + i*step, 4), round(self.start_entry + (i+1)*step, 4)) for i in range(self.n_jobs)}
+
+  def output(self):
+    return self.local_target("stdout_stderr_job_{}.txt".format(self.branch))
```
kandrosov, line 35: here you define a single file for stdout and stderr, but in framework.py there are two files and they have different names. Please, try to have a consistent logging structure.

Production/scripts/ShuffleMergeLaw/analysis/tasks.py
```diff
@@ -0,0 +1,64 @@
...
+      '--input-spec'          , str(self.input_spec)                ,
+      '--tau-ratio'           , quote(str(self.tau_ratio))          ,
+      '--exp-disbalance'      , str(self.exp_disbalance)            ,
+      '--start-entry'         , str(self.branch_data[0])            ,
+      '--end-entry'           , str(self.branch_data[1])            ,
+    ])
+    print ('>> {}'.format(command))
+    
+    proc = subprocess.Popen(command, shell = True, stdout = subprocess.PIPE, stderr = subprocess.PIPE)
+    stdout, stderr = proc.communicate()
```
kandrosov, line 61: - you need to check that the process ended successfully (e.g. that error code is 0). If not, the job should be considered as failed.
- job output root file first should be stored with some tmp name (in sub-directory of the output). Once the job is successfully finished, it should be move to its final location.

Production/scripts/ShuffleMergeLaw/setup.sh
```diff
@@ -0,0 +1,18 @@
...
+    local this_file="$( [ ! -z "$ZSH_VERSION" ] && echo "${(%):-%x}" || echo "${BASH_SOURCE[0]}" )"
+    local this_dir="$( cd "$( dirname "$this_file" )" && pwd )"
+
+    export PYTHONPATH="$this_dir:$PYTHONPATH"
+    export LAW_HOME="$this_dir/.law"
+    export LAW_CONFIG_FILE="$this_dir/law.cfg"
+
+    export ANALYSIS_PATH="$this_dir"
+    export ANALYSIS_DATA_PATH="$ANALYSIS_PATH/data"
+    export SML_INPUT_DATA="/eos/cms/store/group/phys_tau/TauML/prod_2018_v1/full_tuples/WJetsToLNu_HT-1200To2500/"
```
kandrosov, line 13: where `SML_INPUT_DATA` is used?

kandrosov (changes requested):
Additional comment:
- final task to run validation when all shuffle&merge jobs are finished should be defined

shedprog:
After last chenges Shuffle&Merge can be started with this command:
`ShuffleMergeSpectral --cfg <path> --input /eos/cms/store/group/phys_tau/TauML/prod_2018_v1/full_tuples --output <path> --mode MergeAll --n-threads 1 --disabled-branches "trainingWeight" --input-spec /afs/cern.ch/work/m/myshched/public/root-tuple-v2-hists --pt-bins "20, 30, 40, 50, 60, 70, 80, 90, 100, 1000" --eta-bins "0., 0.6, 1.2, 1.8, 2.4" --tau-ratio "jet:1, e:1, mu:1, tau:1, emb_jet:-1, emb_tau:-1" --exp-disbalance 100 --start-entry 0.0 --end-entry 0.001`

I am also propossing to test on a bit smaller but representative data set, if there are problems with the big one:
`
DY: dir=(DY1.*) file=eventTuple.* types=jet,e,mu,tau
TT: dir=(TTTo2L2Nu.*) file=eventTuple.* types=jet,e,mu,tau
W:  dir=(W1Jets.*) file=eventTuple.* types=jet,e,mu,tau
QCD: dir=(QCD_HT700to1000.*) file=eventTuple.* types=jet
Higgs: dir=(ttHTo.*) file=eventTuple.* types=tau
Zprime: dir=(ZprimeToTauTau_M-4000.*) file=eventTuple.* types=e,mu,tau
Emb: dir=(Embedded_MuTau_Run2018A.*) file=eventTuple.* types=emb_tau,emb_jet
`

Production/scripts/ShuffleMergeLaw/analysis/framework.py
lguzzi: __init__.py is needed to run the "law index" command, which otherwise fails.

Production/scripts/ShuffleMergeLaw/analysis/tasks.py
lguzzi: yes, the task output is the same of the framework output (which is the stdout of the condor job), so one of the two is redundant. I will remove it from the task.

Production/scripts/ShuffleMergeLaw/analysis/tasks.py
```diff
@@ -78,3 +78,62 @@ def run(self):
     retcode = proc.returncode
     if proc.returncode != 0:
       raise Exception('job {} return code is {}'.format(self.branch, retcode))
+
+class HaddFiles(Task, HTCondorWorkflow, law.LocalWorkflow):
```
kandrosov, line 82: - I think HaddFiles should be part of a separate workflow (e.g. CrabPostProc). We need to do it only once after the production is done and not each time we re-run ShffleMerge.
- If some files are corrupted, hadd still exits with code 0. To catch such cases, you need to search in the log files word "error" or "warning" (case insensitive) and if it is found, consider the job as failed.

shedprog: Do you think adding -k flag (Skip corrupt or non-existent files) would be an option? and then we can parse output files after that to create the list of corrupted files?

kandrosov: Yes, having `-k` flag is a good idea. Like this, in the log file we'll have all corrupted files (if there are more than one). However, in my opinion, if there is at least one corrupted file, the condor job should be considered as failed and after the problem is solved resubmitted again. I.e. if HaddFiles is done just after the crab production, the corresponding crab job can be resubmitted. If not, one can decide to remove the corrupted files.

Production/scripts/ShuffleMergeLaw/analysis/framework.py
lguzzi: I moved the project to Analysis/law 9a67a29beff0bc8c5878b9bff139073287a1c33d

Production/scripts/ShuffleMergeLaw/analysis/tasks.py
lguzzi: all the S&M arguments are now in the corresponding law task
https://github.com/lguzzi/TauMLTools/blob/law_tool/Analysis/law/ShuffleMergeSpectral/tasks.py#L16-L35

Production/scripts/ShuffleMergeLaw/analysis/tasks.py
lguzzi: done in c2d446ac8404539d3dbc0362193a56b1a38a760a and updated by later commits

Production/scripts/ShuffleMergeLaw/setup.sh
lguzzi: I removed the variable from the script since it was not used 0df9b48c60af99d663ca23808ba066fca26ff588

lguzzi:
I updated the code for merging:

- the script is now under Analysis/law
- the script creates a temporary "tmp" directory. If the job is succesful, the output file is moved one folder up (i.e. in the --output-path)
- I split Hadd and ShuffleMerge tasks into two different files
- I updated the README

What's missing:
- Hadd does not handle corrupted files yet (hadd is not needed by S&M)
- The validation part after S&M is not implemeted. Anyway, this is a single-job task which can be run interactively for now
- The S&M task only works with --mode MergeAll. The other mode is MergePerEntry

I think the tool can be merged and the three bullets above can be adressed later on.

kandrosov (approved):

kandrosov:
Thank you @lguzzi. All looks good.

PR #35, opened 2021-02-08
c++-based DataLoader 

shedprog:
It is preliminary pull request. "[To Do list](https://github.com/shedprog/TauMLTools/projects/1)"

To run tests:
```sh
source /cvmfs/sft.cern.ch/lcg/views/setupViews.sh LCG_99 x86_64-centos7-gcc10-opt
python DataLoader_test.py
``` 
P.S: The run the test, the following lines should be disabled: https://github.com/shedprog/TauMLTools/blob/7d6d92155c0f9d7564dbcb3dd20239e2729d9771/Analysis/interface/TauTuple.h#L7
https://github.com/shedprog/TauMLTools/blob/7d6d92155c0f9d7564dbcb3dd20239e2729d9771/Analysis/interface/TauTuple.h#L122

Current benchmarks are:
for 300 taus -> 1.0 s/batch
for 500 taus -> 1.6 s/batch
for 1k.  taus -> 3.3 s/batch

Updated benchmark of the final version:
for 500 taus -> ~1 s/batch

Time per load of 500 taus/batch:
time

Memory usage for 500 taus/batch:
n_batches = 1000
n_batches in memory = 10 (max queue length)
Mem


Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,770 @@
+#include <boost/preprocessor/variadic.hpp>
+#include <boost/math/constants/constants.hpp>
+
+#include "TauMLTools/Core/interface/AnalysisMath.h"
+#include "TauMLTools/Core/interface/RootExt.h"
+#include "TauMLTools/Analysis/interface/TauTuple.h"
+
+#include "TROOT.h"
+#include "TLorentzVector.h"
+#include "DataLoader_setup.h"
```
kandrosov, line 10: It will be more flexible if the setup is not included directly in `DataLoader_main.h`. Like this, we'll have a possibility to have different setups without changing the DataLoader code.

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,770 @@
+#include <boost/preprocessor/variadic.hpp>
+#include <boost/math/constants/constants.hpp>
+
+#include "TauMLTools/Core/interface/AnalysisMath.h"
+#include "TauMLTools/Core/interface/RootExt.h"
```
kandrosov, line 5: why `RootExt.h` is needed?

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,770 @@
...
+        n_pf_chHad(setup::n_pf_chHad), n_pf_nHad(setup::n_pf_nHad) ,n_pf_gamma(setup::n_pf_gamma),
+        n_electrons(setup::n_ele), n_muons(setup::n_muon), n_labels(setup::tau_types),
+        // trainingWeightFactor(tauTuple.GetEntries() / training_weight_factor,
+        innerCellGridRef(n_inner_cells, n_inner_cells, inner_cell_size, inner_cell_size),
+        outerCellGridRef(n_outer_cells, n_outer_cells, outer_cell_size, outer_cell_size)
+    {
+      if(n_threads > 1) ROOT::EnableImplicitMT(n_threads);
+    }
+
+    void Initialize(std::string file_name)
```
kandrosov, line 161: Can initialization be done in the constructor?

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,770 @@
...
+                                           n_muons,
+                                           n_labels
+                                          );
+
+        const Long64_t end_entry = current_entry + n_tau;
+        size_t n_processed = 0, n_total = static_cast<size_t>(end_entry - start_dataset);
+        for(Long64_t tau_i = 0; tau_i < n_tau; ++tau_i, ++current_entry) {
+            tauTuple->GetEntry(current_entry);
+            const auto& tau = tauTuple->data();
+            if(parity == -1 || tau.evt % 2 == static_cast<unsigned>(parity)) {
```
kandrosov, line 195: We have training/testing split at the level of shuffle&merge. Here it is redundant. Also, the current implementation is wrong because it would not load n_tau in Data if parity != -1.

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,770 @@
...
+      }
+
+
+      void FillTauBranches(const Tau& tau, Long64_t tau_i, std::shared_ptr<Data>& data)
+      {
+        Long64_t start_array_index = tau_i * n_fe_tau;
+
+        // Filling Tau Branche
+        auto get_tau_branch = [&](TauFlat_f _fe) -> float& {
+            size_t _fe_ind = static_cast<size_t>(_fe);
```
kandrosov, line 228: here and in the similar lambdas below: add check that _fe >=0. Setting some variables to -1, it would provide as with an easy way of disabling them.

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,770 @@
...
+            size_t index = start_array_index + _fe_ind;
+            return data->x_tau.at(index);
+        };
+
+        // filling labels one_hot vector
+        Int_t tau_type = tau.tauType;
+        if(tau_type<=3) data->y_onehot[ tau_i * n_labels + tau_type ] = 1;
+        else if(tau_type>=6) data->y_onehot[ tau_i * n_labels + tau_type - 2 ] = 1;
```
kandrosov, line 236: this way to convert tau_type to y_onehot looks error-prone. For the default training setup, we have only 4 tau_types, all other taus (e.g. from embedded) should be ignored: or by moving to the next tau (better) or by setting the weight to 0.

Training/python/DataLoader_test.py
```diff
@@ -0,0 +1,61 @@
...
+n_pf_el     = 22
+n_pf_mu     = 23
+n_pf_chHad  = 27
+n_pf_nHad   = 7
+n_pf_gamma  = 23
+n_ele       = 37
+n_muon      = 37
+tau_types   = 6
+
+R.gROOT.ProcessLine(".include /afs/cern.ch/user/m/myshched/DeepTau/CMSSW_10_6_17/src")
```
kandrosov, line 22: The path should be dynamically determined.

Training/python/DataLoader_test.py
```diff
@@ -0,0 +1,61 @@
...
+    x_grid_inner1 = np.asarray(data.x_grid[4][1]).reshape((n_tau, n_inner_cells, n_inner_cells, n_pf_gamma))
+    x_grid_inner1 = np.asarray(data.x_grid[5][1]).reshape((n_tau, n_inner_cells, n_inner_cells, n_ele))
+    x_grid_inner1 = np.asarray(data.x_grid[6][1]).reshape((n_tau, n_inner_cells, n_inner_cells, n_muon))
+
+    end = time.time()
+    print("end: ",end-start, ' s.')
+    times.append(end-start)
+
+from statistics import mean
+print("Mean time: ", mean(times[1:]))
```
kandrosov, line 61: - what is the reason that times[0] is ignored? Loading the first batch takes much longer wrt to the consecutive?
- we also need a larger scale test with monitoring of memory consumption and the loading speed as a function of time to make sure that we keep memory & speed under control.

kandrosov (changes requested):

Training/interface/DataLoader_main.h
shedprog: I see only the following way to do it:

auto fill_PFelectron = [&](PfCand_electron_f _fe_n, float value) {
size_t _idx = static_cast<size_t>(_fe_n);
if(_idx >= 0){
size_t flat_index = getIndex(n_pf_el, _idx);
return data->x_grid.at(CellObjectType::PfCand_electron).at(inner).at(flat_index);
}
};

and that fill branches like this: fill_PFelectron(Br::pfCand_ele_rel_pt, valid ? tau.pfCand_pt.at(pfCand_idx) / tau.tau_pt : 0);
would you say that it is fine to change everything like this? otherwise I have to allocate additional pointer and return reference to it, I do not see other options.

kandrosov: Why something like this wouldn't work?:
```c++
float foo;

auto get_tau_branch = [&](TauFlat_f _fe) -> float& {
    if(_fe < 0) return foo;
    size_t _fe_ind = static_cast<size_t>(_fe);
    size_t index = start_array_index + _fe_ind;
    return data->x_tau.at(index);
};
```

similarly for the other cases.

Training/interface/DataLoader_setup.h
```diff
@@ -0,0 +1,298 @@
...
+  PfCand_muon,
+  PfCand_chargedHadron,
+  PfCand_neutralHadron,
+  PfCand_gamma,
+  Electron,
+  Muon
+};
+
+enum class TauFlat_f {
+  tau_pt = 0,
```
kandrosov, line 56: e.g. variable list in config will be:
```yaml
Features:
- TauFlat: [ tau_pt, tau_eta, ... ]
```

Training/python/DataLoader_test.py
```diff
@@ -0,0 +1,93 @@
...
+x_inner_6 = np.zeros((n_batches_store, n_tau, n_inner_cells, n_inner_cells, n_ele))
+x_inner_7 = np.zeros((n_batches_store, n_tau, n_inner_cells, n_inner_cells, n_muon))
+
+def to_tensor(data, i):
+
+    y[i] = np.frombuffer(data.y_onehot.data(), dtype=np.float32, count=data.y_onehot.size()).reshape((n_tau, tau_types))
+    w[i] = np.frombuffer(data.weight.data(), dtype=np.float32, count=data.weight.size())
+    x_tau[i] = np.frombuffer(data.x_tau.data(), dtype=np.float32, count=data.x_tau.size()).reshape((n_tau,n_fe_tau))
+
+    x_outer_1[i] = np.frombuffer(data.x_grid[0][0].data(), dtype=np.float32, count=data.x_grid[0][0].size()).reshape((n_tau, n_outer_cells, n_outer_cells, n_pf_el))
```
kandrosov, line 59: as discussed in the meeting, please, check it is possible to use a queue without a pre-defined set of slots.

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,931 @@
...
+            const double cnt_old = old_hist.GetBinContent(bin_old);
+            const double err_old = old_hist.GetBinError(bin_old);
+            const double cnt_new = new_hist.GetBinContent(bin_new);
+            const double err_new = new_hist.GetBinError(bin_new);
+
+            new_hist.SetBinContent(bin_new, cnt_new + cnt_old);
+            new_hist.SetBinError(bin_new, std::hypot(err_new, err_old));
+        }
+    }
+}
```
kandrosov, line 137: To avoid code duplication, can you move all these methods into a separate include and use it also in the main code? I.e. we can rearrange includes in the Core in the way that all methods that are useful at the training step, are implemented in a single .h file.

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,931 @@
...
+
+    size_t GetMaxBatchNumber () {
+        return (end_dataset - start_dataset) / n_tau;
+    }
+
+    std::shared_ptr<Data> LoadNext(){
+
+        if(!tauTuple)
+            throw std::runtime_error("DataLoader is not initialized.");
+
```
kandrosov, line 310: `if(!HasNext())` -> throw an exception

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,931 @@
...
+                                           n_pf_nHad,
+                                           n_pf_gamma,
+                                           n_electrons,
+                                           n_muons,
+                                           n_labels
+                                          );
+
+        const Long64_t end_entry = current_entry + n_tau;
+        size_t n_processed = 0, n_total = static_cast<size_t>(end_entry - start_dataset);
+        for(Long64_t tau_i = 0; tau_i < n_tau; ++current_entry) {
```
kandrosov, line 325: `++current_entry, ++tau_i`, instead of having it on L336

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,931 @@
...
+            tauTuple->GetEntry(current_entry);
+            const auto& tau = tauTuple->data();
+            // skip event if it is not tau_e, tau_mu, tau_jet or tau_h
+            if(tau.tauType < 0 || tau.tauType > 3) continue;
```
kandrosov, line 329: you are skipping events, so
- current implementation of `HasNext()` will not work properly
- control that `current_entry < end_dataset` should be added
- maybe it is easier to create a shuffled tuple without embedded and avoid skipping events?..

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,931 @@
...
+        get_tau_branch(TauFlat_f::neutralIsoPtSumWeightdR03_over_neutralIsoPtSum) =tau.neutralIsoPtSumWeightdR03 / tau.neutralIsoPtSum;
+        get_tau_branch(TauFlat_f::neutralIsoPtSumdR03_over_dR05) = tau.neutralIsoPtSumdR03 / tau.neutralIsoPtSum;
+        get_tau_branch(TauFlat_f::photonPtSumOutsideSignalCone) = tau.photonPtSumOutsideSignalCone;
+        get_tau_branch(TauFlat_f::puCorrPtSum) = tau.puCorrPtSum;
+
+        const bool tau_dxy_valid = std::isnormal(tau.tau_dxy) && tau.tau_dxy > - 10
+                                   && std::isnormal(tau.tau_dxy_error) && tau.tau_dxy_error > 0;
+        get_tau_branch(TauFlat_f::tau_dxy_valid) = tau_dxy_valid;
+        get_tau_branch(TauFlat_f::tau_dxy) =  tau_dxy_valid ? tau.tau_dxy : 0.f;
+        get_tau_branch(TauFlat_f::tau_dxy_sig) =  tau_dxy_valid ? std::abs(tau.tau_dxy)/tau.tau_dxy_error : 0.f;
```
kandrosov, line 406: Use a single if here and in all similar cases below:
```suggestion
if(tau_dxy_valid) {
    get_tau_branch(TauFlat_f::tau_dxy) = tau.tau_dxy;
    get_tau_branch(TauFlat_f::tau_dxy_sig) = std::abs(tau.tau_dxy)/tau.tau_dxy_error;
}
        
        
```

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,931 @@
...
+                                                n_eta_bins, eta_min, eta_max,
+                                                n_pt_bins, pt_min, pt_max);
+        TH2D* input_hist = (TH2D*)file_input->Get(("eta_pt_hist_"+labels_names[type]).c_str());
+
+        RebinAndFill(*hist_weights[type], *target_hist);
+        RebinAndFill(*after_rebin_input_hist, *input_hist);
+        hist_weights[type]->Divide(after_rebin_input_hist);
+        delete after_rebin_input_hist;
+      }
+
```
kandrosov, line 293: Implement a control that `max_weight / min_weight < setup::weight_thr`. If not satisfied -> throw an exception.

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,931 @@
...
+  const size_t n_electrons;
+  const size_t n_muons;
+  const int n_labels;
+  const std::vector<std::string>  labels_names;
+  const int n_eta_bins;
+  const double eta_min;
+  const double eta_max;
+  const int n_pt_bins;
+  const double pt_min;
+  const double pt_max;
```
kandrosov, line 921: these variables are already defined in setup, right? If yes, I don't see the necessity to repeat them here

kandrosov (changes requested):

Training/interface/DataLoader_main.h
kandrosov: alternatively, still, load emb taus but set weight to 0

Training/interface/DataLoader_main.h
shedprog: Unfortunately, here the idea was to increase tau_i counter only when we have needed type (tau.tauType < 0 || tau.tauType > 3) to select exactly n_tau candidates of required type, but if it is not readable at a glimpse I will think how to rearrange it in more elegant manner

Training/interface/DataLoader_main.h
shedprog: > current implementation of HasNext() will not work properly
I agree, but since we are implementing it with a whole chain of files and the percent of emb is negligible I thought it will not arise any problems.

> maybe it is easier to create a shuffled tuple without embedded and avoid skipping events?
It would be definitely and it would retrench the RAM usage for S&M, but in any case I would say DataLoader should be generalized as much as possible. Maybe we can reimplement it in the way that on HasNext() method -> tensor is filled and bool is returned and on LoadNext() filled data is just returned?

kandrosov: Yes, it could be. In such case, method should be renamed to `MoveNext()` and `GetData()` correspondingly. And if `GetData()` is called after `MoveNext()` returned `false`, it should throw an exception.

Training/interface/DataLoader_setup.h
```diff
@@ -0,0 +1,298 @@
+struct setup {
```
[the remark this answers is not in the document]
shedprog: by the generation "on-flight", do you literally mean generate DataLoader_setup.h from python or use other structures that can be fed to c++?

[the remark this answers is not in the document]
kandrosov: yes. It doesn't even need to be a file. ProcessLine will work fine with a string that contains c++ code.

Training/configs/training_v1.yaml
```diff
@@ -0,0 +1,285 @@
...
+    n_threads            :  1
+    n_fe_tau             :  43
+    n_pf_el              :  22
+    n_pf_mu              :  23
+    n_pf_chHad           :  27
+    n_pf_nHad            :  7
+    n_pf_gamma           :  23
+    n_ele                :  37
+    n_muon               :  37
+    tau_types            :  4
```
kandrosov, line 18: tau_types is tau_type_names.size(), right?

Training/configs/training_v1.yaml
```diff
@@ -0,0 +1,285 @@
...
+    outer_cell_size      :  0.05
+    n_threads            :  1
+    n_fe_tau             :  43
+    n_pf_el              :  22
+    n_pf_mu              :  23
+    n_pf_chHad           :  27
+    n_pf_nHad            :  7
+    n_pf_gamma           :  23
+    n_ele                :  37
+    n_muon               :  37
```
kandrosov, line 17: `n_*` variables should be computed automatically based on the list of active features

Training/configs/training_v1.yaml
```diff
@@ -0,0 +1,285 @@
...
+    target_spectrum      : "\"/eos/home-m/myshched/ShuffleSpectrum.root\""
+    n_pt_bins            : 49
+    pt_min               : 20.0
+    pt_max               : 1000.0
+    n_eta_bins           : 4
+    eta_min              : 0.0
+    eta_max              : 2.3
+    weight_thr           : 100000.0
+
+Features_enable :
```
kandrosov, line 34: names (`Features_enable`, `Features_disable`) and the way it is used is a bit counterintuitive for me.
Maybe (`Features_all`, `Features_enabled`) or (`Features_all`, `Features_disabled`) would be better?

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,840 @@
...
+
+      std::cout << "Number of files to process: " << input_files.size() << std::endl;
+      tauTuple = std::make_shared<TauTuple>("taus", input_files);
+      end_entry = std::min((long long)end_dataset, tauTuple->GetEntries());
+
+      // histogram to calculate weights
+      // TH1::AddDirectory(kFALSE);
+      TFile* file_input = new TFile(input_spectrum.c_str());
+      TFile* file_target = new TFile(target_spectrum.c_str());
+      TH2D* target_hist = (TH2D*)file_target->Get("eta_pt_hist_tau");
```
kandrosov, line 163: better to use `dynamic_cast<TH2D*>`

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,840 @@
...
+      // tauTuple = std::make_shared<tau_tuple::TauTuple>(file.get(), true);
+
+      std::cout << "Number of files to process: " << input_files.size() << std::endl;
+      tauTuple = std::make_shared<TauTuple>("taus", input_files);
+      end_entry = std::min((long long)end_dataset, tauTuple->GetEntries());
+
+      // histogram to calculate weights
+      // TH1::AddDirectory(kFALSE);
+      TFile* file_input = new TFile(input_spectrum.c_str());
+      TFile* file_target = new TFile(target_spectrum.c_str());
```
kandrosov, line 162: `
```suggestion
      auto file_input = std::make_shared<TFile>(input_spectrum.c_str());
      auto file_target = std::make_shared<TFile>(target_spectrum.c_str());
```

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,840 @@
...
+      TH2D* target_hist = (TH2D*)file_target->Get("eta_pt_hist_tau");
+
+      for(int type = 0; type < 4; ++type) {
+        hist_weights.push_back(shared_ptr<TH2D>(new TH2D(("w_1_"+tau_types_names[type]).c_str(),
```
kandrosov, line 166: use `make_shared`

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,840 @@
...
+                                                         ("w_1_"+tau_types_names[type]).c_str(),
+                                                         n_eta_bins, eta_min, eta_max,
+                                                         n_pt_bins, pt_min, pt_max)));
+        hist_weights[type]->SetDirectory(0); // disabling the file referencing
+
+        TH2D* after_rebin_input_hist = new TH2D("input_hist", "input_hist",
```
kandrosov, line 172: use `make_shared`

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,840 @@
...
+            FillTauBranches(tau, tau_i, data);
+            FillCellGrid(tau, tau_i, innerCellGridRef, data, true);
+            FillCellGrid(tau, tau_i, outerCellGridRef, data, false);
+            ++tau_i;
+          }
+        }
+        hasData = true;
+        return true;
+    }
+
```
kandrosov, line 216: We also need the method `Reset` that sets current_entry to 0. It will be used to run training for more than one epoch.

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,840 @@
...
+    size_t GetEntries () { return end_entry - current_entry; }
+
+    std::shared_ptr<Data> LoadData() {
+      if(!hasData)
+        throw std::runtime_error("Data was not loaded with MoveNext()");
+      hasData = false;
+      return data;
+    }
+
+    static void MaxDisbCheck(std::vector<std::shared_ptr<TH2D>> hists, Double_t max_thr) {
```
kandrosov, line 226:
```suggestion
    static void MaxDisbCheck(const std::vector<std::shared_ptr<TH2D>>& hists, Double_t max_thr) {
```

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,840 @@
...
+                << ", imbalance threshold: " <<  max_thr << std::endl;
+      if(max_weight / min_weight > max_thr)
+        throw std::runtime_error("The imbalance in the weights exceeds the threshold.");
+    }
+
+  private:
+
+      static constexpr float pi = boost::math::constants::pi<float>();
+
+      const double GetWeight(const int type_id, const double pt, const double eta) const
```
kandrosov, line 243:
```suggestion
      double GetWeight(int type_id, double pt, double eta) const
```

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,840 @@
...
+
+          auto iter = obj_types.find(std::abs(pdgId));
+          if(iter == obj_types.end())
+              throw std::runtime_error("Unknown object pdg id = "+std::to_string(pdgId));
+          return iter->second;
+      }
+
+      CellGrid CreateCellGrid(const Tau& tau, const CellGrid& cellGridRef, bool inner) const
+      {
+          static constexpr double iso_cone = 0.5;
```
kandrosov, line 790: size of the iso cone should be moved to the config

Training/interface/DataLoader_setup.h
```diff
@@ -0,0 +1,288 @@
+struct Setup {
```
kandrosov, line 1: Remove this file.

Training/python/DataLoader_test.py
```diff
@@ -0,0 +1,112 @@
...
+tau_types   = R.Setup.tau_types
+
+data_loader = R.DataLoader()
+print("Number of all entries / batch size: ", data_loader.GetEntries()/n_tau)
+
+n_batches = 1000
+n_batches_store = 5
+
+# To convert list to tensor
+def to_tensor(data_list):
```
kandrosov, line 35: to_tensor should transform only 1 item no list

Training/python/DataLoader_test.py
```diff
@@ -0,0 +1,112 @@
...
+times = []
+for i in range(n_batches):
+    start = time.time()
+
+    data_loader.MoveNext()
+    data = data_loader.LoadData()
+
+    if data_que.full():
+        data_que.get()
+    data_que.put(data)
```
kandrosov, line 98: To representative time measurements `to_tensor` should be called here:
```python
data_np = to_tensor(data)
data_que.put((data, data_np))
```

Training/python/yaml_parse.py
```diff
@@ -0,0 +1,38 @@
+import yaml
+
+def create_namestruc(name: str, content: dict) -> str:
+    types_map = {
+        int   : "size_t",
+        float : "Double_t",
+        str   : "std::string",
+        dict  : "std::vector<std::string>"
```
kandrosov, line 8: use `list` instead `dict` to guarantee the order preservation

kandrosov (changes requested):

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,868 @@
+#include "TauMLTools/Core/interface/AnalysisMath.h"
+#include "TauMLTools/Analysis/interface/TauTuple.h"
+#include "TauMLTools/Training/interface/DataLoader_tools.h"
+
+#include "TROOT.h"
+#include "TLorentzVector.h"
+
+using namespace std;
+
+enum class CellObjectType {
```
kandrosov, line 10: `CellObjectType` should be defined in setup.h. When generating setup, you need additionally generate helper class for features:
```c++
template<typename T>
struct FeaturesHelper;

template<>
FeaturesHelper<PfCand_electron_Features> {
    static constexpr CellObjectType object_type = CellObjectType::PfCand_electron;
    static constexpr size_t size = 2;
    using scaler_type = Scaling::PfCand_electron;
};

...
```

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,868 @@
...
+  PfCand_electron,
+  PfCand_muon,
+  PfCand_chargedHadron,
+  PfCand_neutralHadron,
+  PfCand_gamma,
+  Electron,
+  Muon
+};
+
```
kandrosov, line 19: I propose add definition of the following types:
```c++
template <typename T, typename Tuple>
struct ElementIndex;

template <typename T, typename... Args>
struct ElementIndex<T, std::tuple<T, Args...>> {
    static constexpr std::size_t value = 0;
};

template <typename T, typename U, typename... Args>
struct ElementIndex<T, std::tuple<U, Args...>> {
    static constexpr std::size_t value = 1 + ElementIndex<T, std::tuple<Args...>>::value;
};

using FeatureTuple = std::tuple<PfCand_electron_Features,
                                PfCand_muon_Features,
                                PfCand_chHad_Features,
                                PfCand_nHad_Features,
                                pfCand_gamma_Features,
                                Electron_Features,
                                Muon_Features>;

```

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,868 @@
...
+                 + cellGridRef.GetFlatIndex(cellIndex) * n_total;
+        };
+
+        size_t PfCand_electron_idx_start = getStartIndex(n_PfCand_electron);
+        size_t PfCand_muon_idx_start = getStartIndex(n_PfCand_muon);
+        size_t PfCand_chHad_idx_start = getStartIndex(n_PfCand_chHad);
+        size_t PfCand_nHad_idx_start = getStartIndex(n_PfCand_nHad);
+        size_t pfCand_gamma_idx_start = getStartIndex(n_pfCand_gamma);
+        size_t Electron_idx_start = getStartIndex(n_Electron);
+        size_t Muon_idx_start = getStartIndex(n_Muon);
```
kandrosov, line 402: Define CreateStartIndices as a function:
```c++
template<size_t... I>
std::vector<size_t> CreateStartIndices(const CellGrid& grid, const CellIndex& cellIndex, std::index_sequence<I...>)
{
     auto getStartIndex = [&](size_t n_total) {
          return tau_i * cellGridRef.GetnTotal() * n_total
                 + cellGridRef.GetFlatIndex(cellIndex) * n_total;
    };
    std::vector<size_t> start(nFeaturesTypes);
    ((start[I] = getStartIndex(FeaturesHelper<std::tuple_element_t(I, FeatureTuple)>::size)), ...);
    return start;
}
```
and then call it here:
```c++
static constexpr size_t nFeaturesTypes = std::tuple_size_v<FeatureTuple>;

const auto start_indices = CreateStartIndices(cellGridRef, cellIndex, std::make_index_sequence<nFeaturesTypes>{});
```

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,868 @@
...
+
+        // to be cahnged in c++20 version (auto fillGrid = []<typename ...Ts>(Ts&& ...ts) ...)
+        // decltype(_feature_idx) -> Ts
+        auto fillGrid = [&](auto _feature_idx, float value) -> void {
```
kandrosov, line 406: Then fillGrid will be something like this:
```c++
auto fillGrid = [&](auto _feature_idx, float value) {
    if(static_cast<int>(_feature_idx) < 0) return;
    const CellObjectType obj_type = FeaturesHelper<decltype(_feature_idx)>::object_type;
    const size_t start = start_indices.at(ElementIndex<decltype(_feature_idx), FeatureTuple>::value);

    data->x_grid.at(obj_type).at(inner).at(start + static_cast<int>(_feature_idx))
            = Scale<FeaturesHelper<decltype(_feature_idx)>::scaler_type>(static_cast<int> (_feature_idx), value, inner);
};
```

Training/python/DataLoader_test.py
```diff
@@ -0,0 +1,120 @@
...
+    x_inner_4 = np.zeros((n_tau, n_inner_cells, n_inner_cells, n_pf_nHad))
+    x_inner_5 = np.zeros((n_tau, n_inner_cells, n_inner_cells, n_pf_gamma))
+    x_inner_6 = np.zeros((n_tau, n_inner_cells, n_inner_cells, n_ele))
+    x_inner_7 = np.zeros((n_tau, n_inner_cells, n_inner_cells, n_muon))
+
+    y = np.frombuffer(data.y_onehot.data(), dtype=np.float32, count=data.y_onehot.size()).reshape((n_tau, tau_types))
+    w = np.frombuffer(data.weight.data(), dtype=np.float32, count=data.weight.size())
+    x_tau = np.frombuffer(data.x_tau.data(), dtype=np.float32, count=data.x_tau.size()).reshape((n_tau,n_fe_tau))
+
+    x_outer_1 = np.frombuffer(data.x_grid[0][0].data(), dtype=np.float32, count=data.x_grid[0][0].size()).reshape((n_tau, n_outer_cells, n_outer_cells, n_pf_el))
```
kandrosov, line 66: - if we use frombuffer, do we still need to initialize with zeros above?
- convert to return lists

Training/interface/DataLoader_main.h
```diff
@@ -0,0 +1,868 @@
...
+      // tauTuple = std::make_shared<tau_tuple::TauTuple>(file.get(), true);
+
+      std::cout << "Number of files to process: " << input_files.size() << std::endl;
+      tauTuple = std::make_shared<TauTuple>("taus", input_files);
+      end_entry = std::min((long long)end_dataset, tauTuple->GetEntries());
+
+      // histogram to calculate weights
+      auto file_input = std::make_shared<TFile>(input_spectrum.c_str());
+      auto file_target = std::make_shared<TFile>(target_spectrum.c_str());
+      std::shared_ptr<TH2D> target_hist(dynamic_cast<TH2D*>(file_target->Get("eta_pt_hist_tau")));
```
kandrosov, line 164: here and below, check that histogram is not nullptr before using it.

kandrosov (changes requested):

Training/interface/DataLoader_main.h
shedprog: I am a bit struggling with this inside .gInterpreter.ProcessLine() for some reasons

kandrosov: and if you generate a temporary .h file and then `ProcessLine('#include "tmp.h"')`?

shedprog: With the temporary header it is working

Training/interface/DataLoader_main.h
```diff
@@ -5,18 +5,26 @@
...
+struct ElementIndex<T, std::tuple<T, Args...>> {
+    static constexpr std::size_t value = 0;
+};
+
+template <typename T, typename U, typename... Args>
+struct ElementIndex<T, std::tuple<U, Args...>> {
+    static constexpr std::size_t value = 1 + ElementIndex<T, std::tuple<Args...>>::value;
 };
 
+using FeatureTuple = std::tuple<PfCand_electron_Features,
```
kandrosov, line 21: perhaps `using FeatureTuple` should be generated in python to be consistent with the order in `CellObjectType`.

Training/interface/DataLoader_main.h
```diff
@@ -751,8 +730,8 @@ class DataLoader {
             fillGrid(Br::muon_normalizedChi2_valid, static_cast<float>(normalizedChi2_valid));
 
             if(normalizedChi2_valid){
-              fillGrid(Br::muon_normalizedChi2, normalizedChi2_valid ? tau.muon_normalizedChi2.at(idx) : 0);
-              fillGrid(Br::muon_numberOfValidHits, normalizedChi2_valid ? tau.muon_numberOfValidHits.at(idx) : 0);
+              fillGrid(Br::muon_normalizedChi2, tau.muon_normalizedChi2.at(idx)); //goes to Inf in some very rare cases
```
kandrosov, line 733: add protection that `muon_normalizedChi2` is finite

kandrosov (changes requested):

kandrosov:
@shedprog all looks good. Thank you! Could you, please, update the benchmark values in the first PR message to reflect the situation with the latest version of the code? It can be useful for future references.

kandrosov (approved):

shedprog:
> It is preliminary pull request. "[To Do list](https://github.com/shedprog/TauMLTools/projects/1)"
> 
> To run tests:
> 
> ```shell
> source /cvmfs/sft.cern.ch/lcg/views/setupViews.sh LCG_99 x86_64-centos7-gcc10-opt
> python DataLoader_test.py
> ```
> 
> Current benchmarks are:
> for 300 taus -> 1.0 s/batch
> for 500 taus -> 1.6 s/batch
> for 1k. taus -> 3.3 s/batch
> 
> Updated benchmark of the final version:
> for 500 taus -> ~1 s/batch
> 
> Time per load of 500 taus/batch:
> time
> 
> Memory usage for 500 taus/batch:
> n_batches = 1000
> n_batches in memory = 10 (max queue length)
> Mem

This initial description is updated

PR #40, opened 2021-03-19
gen cuts in ShuffleMergeSpectral plus minor modifications

lguzzi:
This PR updates the Analysis/bin/ShuffleMergeSpectral.cxx and Analysis/bin/CreateSprectralHists.cxx scripts with cuts at gen level, based on the [HTT definitions](https://twiki.cern.ch/twiki/bin/viewauth/CMS/HiggsToTauTauWorking2016#MC_Matching).

It also enables the build of Analysis/bin scripts in this branch and changes the name of lepton_gen_match to genLepton_kind where needed to match the definition of Analysis/interface/TauTuple.h. 

Minor changes:
- removed some unused files in Analysis/bin

  -  Analysis/bin/CreateBalancedTuple.cxx
  -  Analysis/bin/CreateBinnedTuples.cxx
  -  Analysis/bin/ShuffleMerge.cxx
  -  Analysis/bin/ShuffleTuple.cxx 
  - Analysis/bin/ShuffleTupleEntries.cxx 
  - Analysis/interface/TrainingTuple.h 
  - Analysis/bin/TrainingTupleProducer.cxx
  
- changed the behaviour of [GenMatchToTauType](https://github.com/cms-tau-pog/TauMLTools/blob/master/Analysis/interface/AnalysisTypes.h#L65) to include events with genLepton_index <0 and genJet_index >= 0
- fixed [comparison of floats](https://github.com/cms-tau-pog/TauMLTools/blob/master/Analysis/bin/CreateSpectralHists.cxx#L115-L119) (was comparing strings) 
- removed MergePerEntry from possible merging modes of ShuffleMergeSpectral.cxx since it was broken and never used


kandrosov:
I suggest removing all outdated c++ tools from Analysis/bin and keeping only: ShuffleMergeSpectral, CreateSpectralHists and MergeTuples.

Analysis/bin/BuildFile.xml
```diff
@@ -31,9 +30,3 @@
     <use name="boost_program_options"/>
     <use name="TauMLTools/Core"/>
 </bin>
```
kandrosov, line 32: remove references to the old code in BuildFile.xml.

Analysis/bin/CreateSpectralHists.cxx
```diff
@@ -126,9 +130,9 @@ class CreateSpectralHists {
     void AddTau(const Tau& tau)
     {
         if(PassSelection(tau)) {
-            const auto gen_match = static_cast<analysis::GenLeptonMatch>(tau.lepton_gen_match);
+            const auto gen_match = static_cast<analysis::GenLeptonMatch>(tau.genLepton_kind);
```
kandrosov, line 133: Type of `genLepton_kind` is `reco_tau::gen_truth::GenLepton:: Kind`. Don't cast it to `analysis::GenLeptonMatch`, because in the future these definitions can diverge.

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -16,24 +16,24 @@
...
     run::Argument<std::string> input{"input", "input path with tuples for all the samples"};
-    run::Argument<std::string> output{"output", "output, depending on the merging mode: MergeAll - file,"
-                                                " MergePerEntry - directory."};
+    run::Argument<std::string> output{"output", "output, depending on the merging mode: MergeAll (by file) "
+                                                "is the only supported mode"};
     run::Argument<std::string> pt_bins{"pt-bins", "pt bins (last bin will be chosen as high-pt region, \
                                                    lower pt edgeof the last bin will be choosen as pt_threshold)"};
     run::Argument<std::string> eta_bins{"eta-bins", "eta bins"};
-    run::Argument<MergeMode> mode{"mode", "merging mode: MergeAll or MergePerEntry"};
+    run::Argument<MergeMode> mode{"mode", "merging mode: MergeAll is the only supported mode"};
```
kandrosov, line 36: set value by default, since it is the only supported mode...

Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -105,9 +105,13 @@ struct SourceDesc {
               throw exception("Root file %1% is empty.") % file_name;
         }
         current_tuple->GetEntry(current_n_processed++);
-        const auto gen_match = static_cast<GenLeptonMatch>((*current_tuple)().lepton_gen_match);
+
+        if (!PassGenLeptonCut((*current_tuple)())) return false;
```
kandrosov, line 109: I think it should be `continue` instead `return false`.

Analysis/interface/TauSelection.h
```diff
@@ -0,0 +1,43 @@
...
+            }
+        } 
+        else if (tau.genLepton_index < 0 && tau.genJet_index >= 0){
+            return true;
+        } 
+        else {
+            return false;
+        }
+        throw exception("Invalid execution. tau.genLepton_index = %1% tau.genLepton_kind = %2%")
+            %tau.genLepton_index %tau.genLepton_kind;
```
kandrosov, line 41:
```suggestion
        return tau.genJet_index >= 0;
```
or with the proposed updated signature:
```c++
boost::optional<GenLeptonMatch> match;
if(tau.genLepton_index >= 0) {
    ...
} else if(tau.genJet_index >= 0) {
    match = GenLeptonMatch::NoMatch;
}
return match;
```

Analysis/interface/AnalysisTypes.h
```diff
@@ -62,8 +62,12 @@ ENUM_NAMES(SampleType) = {
     { SampleType::Data, "Data" }, { SampleType::MC, "MC" }, { SampleType::Embedded, "Embedded" }
 };
 
-inline TauType GenMatchToTauType(GenLeptonMatch gen_match, SampleType sample_type)
+inline TauType GenMatchToTauType(GenLeptonMatch gen_match, SampleType sample_type, const int genLepton_index, const int genJet_index)
```
kandrosov, line 65: I suggest keeping the `GenMatchToTauType` definition as it was. And return `boost::optional<GenLeptonMatch>` in `PassGenLeptonCut`

Analysis/interface/TauSelection.h
```diff
@@ -0,0 +1,43 @@
+#include "TauMLTools/Analysis/interface/TauTuple.h"
+#include "TauMLTools/Core/interface/exception.h"
+
+#include "TLorentzVector.h"
+
+namespace analysis {
+    bool PassGenLeptonCut(const tau_tuple::Tau& tau)
+    {   
+        if (tau.genLepton_index >= 0){
+            TLorentzVector genv;
```
kandrosov, line 10: use analysis::LorentzVectorM

Analysis/interface/TauSelection.h
```diff
@@ -0,0 +1,43 @@
+#include "TauMLTools/Analysis/interface/TauTuple.h"
+#include "TauMLTools/Core/interface/exception.h"
+
+#include "TLorentzVector.h"
+
+namespace analysis {
+    bool PassGenLeptonCut(const tau_tuple::Tau& tau)
```
kandrosov, line 7: Perhaps `boost::optional<GenLeptonMatch> GetGenLeptonMatch(const tau_tuple::Tau& tau)` would be better

Analysis/interface/TauSelection.h
```diff
@@ -0,0 +1,43 @@
...
+            genv.SetPtEtaPhiM(  tau.genLepton_vis_pt   ,
+                                tau.genLepton_vis_eta  ,
+                                tau.genLepton_vis_phi  ,
+                                tau.genLepton_vis_mass );
+            
+            if (tauv.DeltaR(genv) > 0.2){
+                return false;
+            }
+
+            if (tau.genLepton_kind > 0 && tau.genLepton_kind < 5){
```
kandrosov, line 23: To be more robust in case of possible changes of the definitions, better to cast `genLepton_kind` to `reco_tau::gen_truth::GenLepton::Kind` and explicitly compare enum values.

kandrosov (changes requested):

shedprog:
Maybe also in SourceDesc we should not using ShuffleFiles() now, I did it for additional "mixing" but let's better remove it and maybe do it explicitly at pre-processing step with files (if needed). Especially if we will modify SourceDesc for dividing files within jobs.
https://github.com/lguzzi/TauMLTools/blob/8f5128fdb2bf2d824ef8c15c89440a2b5994d040/Analysis/bin/ShuffleMergeSpectral.cxx#L128-L138

Analysis/interface/TauSelection.h
```diff
@@ -0,0 +1,60 @@
...
+                                tau.genLepton_vis_phi ,
+                                tau.genLepton_vis_mass);
+
+            if (ROOT::Math::VectorUtil::DeltaR(tauv, genv) > 0.2){
+                return boost::none;
+            }
+
+            if (genLepton_kind == GTGenLeptonKind::PromptElectron){
+                if (tau.genLepton_vis_pt < 8.0) return boost::none;
+                match = GenLeptonMatch::Electron;
```
kandrosov, line 28: you are mixing 'match = ...' and `return`. Maybe better always use `return`  and remove `GenLeptonMatch match`?

kandrosov (changes requested):

Analysis/interface/TauSelection.h
lguzzi: done in https://github.com/cms-tau-pog/TauMLTools/pull/40/commits/7762e783ce87d67576c9c4e6f8f2b957b57f0858

kandrosov (approved):

PR #41, opened 2021-03-30
bugfix in law and ShuffleMergeSpectral.cxx

lguzzi:
This PR fixes different bugs in ShuffleMergeSpectral.cxx and its law task.

- file_names and name_hashes is not filled in ShuffleMergeSpectral.cxx following my latest PR.
thanks to @shedprog for pointing this out
https://github.com/cms-tau-pog/TauMLTools/pull/40#discussion_r603506344
- now check for spectral files to exist on disk
- fix a tautype bug 9330cd30bfdf14b7c626a770a9f1cac29a08eed4
- throw exception if a wanted tau type has no spectral histogram inside the spectrum file
- throw an exception if a tau type is specified in --tau-ratio but not in the input cfg
- fix law not overwriting existing files, thus failing
- fix some bugs in the S&M law task


Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -525,27 +529,52 @@ class DataSetProcessor {
...
+        }
+      }
+
       for(auto tauR_: tau_ratio){
         if(tauR_.second!=-1){
           if(accumulated_entries.at(tauR_.first)==0)
             throw exception("No taus of the type '%1%' are found in the tuples") % tauR_.first;
+          else if(tauR_.second<0)
+            throw exception("Available --tau_ratio arguments should be > 0 or -1");
+
```
shedprog, line 551: Shouldn't we add one more fix here that you proposed:
if ( accumulated_entries.find(tauR_.first) == accumulated_entries.end())

lguzzi: it's done here https://github.com/cms-tau-pog/TauMLTools/pull/41/files#diff-d1b44f30826c34ea73d4cd6c0df5d77d0387b7b7f0af33361734fbd61706bdd2R537-R544 for both tauR_.secoond ==-1 and != -1

shedprog (approved):

PR #44, opened 2021-04-21
Update of Shuffle&Merge parallelization strategy 

shedprog:
In the following pull request changes of the parallelization and the pt-eta spectrum smoothing are introduced:

1. CreateSpectralHists.cxx and CreateSpectralHists.py are updated to create text files with the data files names and number of the entries.
2. ShuffleMergeSpectral.cxx is updated to parallelize input data files into --n-jobs. (Option of parallelization by 
3. Option to take into account last pt-bin when calculating the probability of rejecting events by pt-eta is added (in addition to the option to take all events from the last pt-bin of the histogram)
4. README is updated to include corresponding options.


kandrosov (approved):
Thank you @shedprog ! The implementation reflects what was discussed during the last meeting. I don't see any obvious bugs.

PR #47, opened 2021-05-26
Histogram_2D class

lguzzi:
This PR add the Histogram_2D class. This class implements custom 2D histograms where the y binning depends on the x bin. The class is used in the dataloader c++ library.

Note: in the current implementation, x = eta, y = pt


Training/configs/training_v1.yaml
```diff
@@ -23,6 +23,12 @@ Setup:
     eta_min              : 0.0
     eta_max              : 2.3
     weight_thr           : 100000.0
+    # here define variables for the Histogram_2D class
+    xaxis_vec            : [0., 0.625, 1.25, 1.875, 2.5]
+    yaxis_list_vec       : [[20., 40., 60., 100., 1000.], [20., 60., 100., 1000.],
+                            [20., 100., 1000.]          , [20., 100., 1000.]     ]
+    ymin_val             : 20.0
+    ymax_val             : 1000.
```
shedprog, line 32: @lguzzi There might be a problem with such approach:
On practice, I think we want to save high granularity for the region of (low pt and high eta), because we have very good statistic there, and there is no reason of not to equalize the spectrum in a tight manner. So if we want to have something like this:
image
Will we be able to do this? because that what we actually want on practice. Or I am missing something? @kandrosov

lguzzi: not exactly. What you could do is split the big bin over x and maybe use only two y bins in the last x bin

kandrosov: uh... I think @shedprog is right, we've overlooked that with the current implementation the division as illustrated above is not possible... Actually, I think we can use [TH2Poly](https://root.cern.ch/doc/master/classTH2Poly.html) as an internal container in `Histogram_2D` class - it would allow us to have binning as above.

shedprog (changes requested):

Training/configs/training_v1.yaml
lguzzi: I've updated the code and switched the x and y axes (the names of the methods and variables are the same though). 
The only modification to be done the code is now to switch x and y in the yaml file. Please note that the yaml x-binning is now the actual y binning of the input histogram, and y-binnings are now the x binnings of the input histogram; e.g. add_y_binning_by_index actually defines the binning over x for a y bin.
I've marked the code where this workaround has been introduced by a comment "// WORKAROUND" and the old code is still present and commented.
Now the code produces results like:
immagine

shedprog: @ Is it some tesing histogram, because in our MC we have eta only up to 2.5 and here it is up to 5.0?

Training/interface/DataLoader_main.h
```diff
@@ -147,10 +148,15 @@ class DataLoader {
         outerCellGridRef(n_outer_cells, n_outer_cells, outer_cell_size, outer_cell_size),
         input_files(FindInputFiles(input_dirs,file_name_pattern,
                                    exclude_list, exclude_dir_list)),
-        hasData(false)
+        hasData(false),
+        xaxis(xaxis_vec), yaxis_list(yaxis_list_vec), ymin(ymin_val), ymax(ymax_val)
```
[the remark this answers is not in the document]
lguzzi: I removed these unneeded variables and updated the code

Training/configs/training_v1.yaml
lguzzi: yes, this is a test. The upper TH2 is actually the input for the test

Training/interface/DataLoader_main.h
```diff
@@ -161,28 +167,34 @@ class DataLoader {
...
-      std::shared_ptr<TH2D> target_hist(dynamic_cast<TH2D*>(file_target->Get("eta_pt_hist_tau")));
 
+      Histogram_2D target_histogram("target", xaxis, ymin, ymax);
+      Histogram_2D input_histogram ("input" , xaxis, ymin, ymax);
+      for (int i = 0; i < yaxis_list.size(); i++){
+          target_histogram.add_y_binning_by_index(i, yaxis_list[i]);
+          input_histogram .add_y_binning_by_index(i, yaxis_list[i]);
+      }
+
+      auto target_th2d = dynamic_cast<TH2D*>(file_target->Get("eta_pt_hist_tau"));
```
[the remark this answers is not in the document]
lguzzi: I updated the code to use shared pointers

kandrosov (approved):

PR #49, opened 2021-07-06
Allow subranges in Histogram2D

lguzzi:
This PR allows to use subranges on x and y in the custon histogram 2D class


lguzzi:
validation: https://indico.cern.ch/event/1056615/contributions/4440540/attachments/2276712/3867822/update_histogram2d.pdf

PR #52, opened 2021-08-03
update dataloader to compute tau type

lguzzi:
This PR updates the Dataloader library to recompute the tau type. The recomputation is controlled by a boolean in the yaml cfg file


shedprog:
hm, @lguzzi now after implementation of tau_type recalculation we have a lot of warnings and potentially error due to the exception.h linking, maybe that should be solved in the future:
image

PR #56, opened 2021-08-11
Fix of few DataLoader interface bugs

shedprog:
1) fix bug with n_batches=-1 case
2) add opportunity to run with validation_split=0 to enable easy prediction evaluation on fixed datafiles
3) *there was bug with input queue_files in the case when we have a lot of files, but I can not reproduce this bug at the moment. Any ideas? (buffer error expected)

thanks @yaourtpourtoi for reporting about the bugs.


kandrosov (changes requested):

shedprog:
@kandrosov done. Bug-3 is not valid anymore it was caused by bug-1. This PR can be merged.

PR #57, opened 2021-08-12
Merge exception.h and exception.cpp to enable jit compilation in python

lguzzi:
This PR merges the Core/interface/exception.h and Core/interface/exception.cpp into Core/exception.hpp to enable jit compilation in the python dataloader, removing some compilation errors.


Analysis/interface/TauSelection.h
```diff
@@ -3,7 +3,7 @@
 #include "TauMLTools/Analysis/interface/GenLepton.h"
 #include "TauMLTools/Analysis/interface/AnalysisTypes.h"
 #include "TauMLTools/Core/interface/AnalysisMath.h"
-#include "TauMLTools/Core/interface/exception.h"
+#include "TauMLTools/Core/src/exception.hpp"
```
kandrosov, line 6: please keep the name as `TauMLTools/Core/interface/exception.h`, because including files from src looks unusual. Also like this you don't need to modify all the files where it is used

Core/src/exception.hpp
```diff
@@ -0,0 +1,91 @@
+/*! Definition of the base exception class for the analysis namespace.
+This file is part of https://github.com/hh-italian-group/TauMLTools. */
+
+#pragma once
+
+#include <exception>
+#include <memory>
+#include <sstream>
+#include <boost/format.hpp>
+#include <boost/stacktrace.hpp>
```
kandrosov, line 10: Last time I checked, with stacktrace.hpp the code was not compiling with CMSSW. Please, check that `scram build` works.

kandrosov (changes requested):

Analysis/interface/TauSelection.h
lguzzi: If I keep the file in interface, some other files (e.g. Production/src/GenTruthTools.cc) are missing the definition of analysis::exception. I believe that scram compiles into shared libraries only what is under src/
As a solution, we can leve exception under interface/exception.h and explicitly include exception.h in all these files

kandrosov: if the whole class is declared in .h, there shouldn't be problems with missing definitions. Maybe you need to run `scram b clean`?

lguzzi: indeed, it compiles now

Training/python/DataLoader.py
```diff
@@ -147,6 +147,7 @@ def compile_classes(file_config, file_scaling):
         R.gInterpreter.Declare(config_parse.create_scaling_input(file_scaling,file_config, verbose=False))
         R.gInterpreter.Declare(config_parse.create_settings(file_config, verbose=False))
         R.gInterpreter.Declare('#include "{}"'.format(_LOADPATH))
+        R.gInterpreter.Declare('#include "TauMLTools/Core/src/exception.hpp"')
```
kandrosov, line 150: - should `exception.h` be included explicitly here?
- if yes, fix the path
- same for `DataLoader_test.py`

kandrosov (changes requested):

Training/python/DataLoader.py
lguzzi: fixed

kandrosov (approved):

PR #58, opened 2021-08-31
Minor DataLoader debug fixes

shedprog:


Training/python/2018v1/Training_v0p1.py
```diff
@@ -253,7 +254,7 @@ def run_training(train_suffix, model_name, model, data_loader, is_profile):
 
 
 config   = os.path.abspath( "../../configs/training_v1.yaml")
-scaling  = os.path.abspath("../../configs/scaling_params_v1.json")
+scaling  = os.path.abspath("/nfs/dust/cms/user/mykytaua/dataDeepTau/DeepTauTraining/ShuffleMergeSpectral_TrainingScaling/ShuffleMergeSpectral_trainingSamples-2_files_0_50.json")
```
kandrosov, line 257: if scaling is updated, add it to the repository, don't refer to the files on nfs.

kandrosov (changes requested):
- Add following command line arguments to Training_v0p1.py:
  - cfg: path to yaml config
  - gpu-index: index of GPU to be used (and in the code `tf.config.set_visible_devices(gpus[args.gpu_index], 'GPU')`)
  - gpu-memory (with default value 7 GB)
  - output (path were to store output models, logs etc.)
- set default batch size to 250
- There is some bug with input file processing. Please fix it:
  ```
  Error in <TFile::ReadBuffer>: error reading all requested bytes from file /eos/cms/store/group/phys_tau/TauML/prod_2018_v2/ShuffleMergeSpectral_trainingSamples-2/hashes/out_329/datagroup_hash.json, got 286 of 300
  Error in <TFile::Init>: /eos/cms/store/group/phys_tau/TauML/prod_2018_v2/ShuffleMergeSpectral_trainingSamples-2/hashes/out_329/datagroup_hash.json failed to read the file type data.```

shedprog:
As discussed with Oleg, changes to the input arguments will be introduced in his PR.

Training/python/2018v1/Training_v0p1.py
```diff
@@ -218,10 +205,20 @@ def compile_model(model, learning_rate):
     model.compile(loss=TauLosses.tau_crossentropy_v2, optimizer=opt, metrics=metrics, weighted_metrics=metrics)
 
 
-def run_training(train_suffix, model_name, model, data_loader, is_profile):
+def run_training(train_suffix, model_name, model, data_loader, to_profile):
+
+    gpus = tf.config.list_physical_devices('GPU')
+    if gpus:
+        try:
+            tf.config.experimental.set_virtual_device_configuration(gpus[data_loader.gpu_index],
```
kandrosov, line 213: With my setup the current code still allocates memory in more than one GPU. Please, add `tf.config.set_visible_devices(gpus[data_loader.gpu_index], 'GPU')`

kandrosov (changes requested):
OK for having cmd line arguments in Oleg's PR, but please add to `.yaml` all constants used in the following lines
- `TauLosses.SetSFs(1, 2.5, 5, 1.5)`
- `model_name = "DeepTau2018v0"`
- `compile_model(model, 1e-3)`

kandrosov:
And also all these constants:
```
tau_net_setup = NetSetup('PReLU', None, 0.2, 128, 128, 1.4, None, False)
comp_net_setup = NetSetup('PReLU', [1, 2], 0.2, 1024, 64, 1.6, None, False)
dense_net_setup = NetSetup('PReLU', None, 0.2, 200, 64, 1.4, None, False)
```

kandrosov:
One other thing. `set` do not guarantee the order preservation, which is important for having a reproducible c++ implementation. Please, fix [this](https://github.com/shedprog/TauMLTools/blob/master/Training/python/DataLoader.py#L323-L324) code to have features listed in the original `.yaml` order. 
```python
        def get_branches(config, group):
            return list(
                set(sum( [list(d.keys()) for d in config["Features_all"][group]],[])) - \
                set(config["Features_disable"][group])
            )
```

kandrosov:
> Hm, I am a bit confused I did not think the order was important here, these lists are needed just to obtain the lengths of the tensor if I am not missing anything.

hm... You are right, now indices are generated in a different place. I was confused with how the things were done in the DeepTau v2 version of the code (there the order of features in this vector was important). Sorry!

yaourtpourtoi:
> OK for having cmd line arguments in Oleg's PR, but please add to `.yaml` all constants used in the following lines
> 
> * `TauLosses.SetSFs(1, 2.5, 5, 1.5)`
> * `model_name = "DeepTau2018v0"`
> * `compile_model(model, 1e-3)`

To be honest I have a feeling that the yaml config is growing too large by now... But on the other note, if we add losses/model info to it, I was wondering about bookkeeping & logging of that info in general. That is, what is needed to be stored in `mlflow` about the model and training procedure and how? I already keep the yaml file as an artifact (so the constants above will be stored), but there is nothing explicit about the model itself, loss/metric definitions, optimisation procedure. There is of course the tf summary of the model as txt + some optimisation params (e.g. optimiser type) logged by default by `mlflow`, but is it enough to reconstruct uniquely what happened during that run?

yaourtpourtoi:
> @yaourtpourtoi - a good point. The set of parameters will change all the time, so having a copy of Training.py could be a good idea. Also .yaml for different types of architectures will contain different parameters (e.g. CNN, GNN etc.). I think it is an unavoidable feature of R&D - things evolve constantly.

Sure, and it is also a sign that the DeepTau framework is growing :) I do have a feeling that it can be designed nicely, so that each architecture, defined as a class, comes with its own yaml cfg file with its parameters going directly to the class constructor. Then keeping track of those configs will be sufficient to uniquely identify the architecture. But ok, at the moment logging training.py should do the job.

On the other hand, I just recalled that mlflow logs the git commit under which the code was executed. So another possibility would be to agree that we commit every architecture to git (maybe on a separate branch), but I guess this would become messy in the end. 

> To keep at least some level of organisation we can agree that the list of parameters for the given architecture should generally increase and amount of exceptions should be as few as we can avoid without compromising productivity.

Hmm, I am not sure that I completely understood your point here, what do you mean by exceptions?

kandrosov:
> Sure, and it is also a sign that the DeepTau framework is growing :) I do have a feeling that it can be designed nicely, so that each architecture, defined as a class, comes with its own yaml cfg file with its parameters going directly to the class constructor. Then keeping track of those configs will be sufficient to uniquely identify the architecture. But ok, at the moment logging training.py should do the job.

The problem is that even for the given architecture we'll change the code and the probability that we will continue coming up with something new that we did though a few months ago is 100% :)

> On the other hand, I just recalled that mlflow logs the git commit under which the code was executed. So another possibility would be to agree that we commit every architecture to git (maybe on a separate branch), but I guess this would become messy in the end.

Yes, I also think that it would not decrease the entropy by much. Let's try to have most of the meta-information in .yaml, e.g. each time when we add something to the code it is reflected in a new parameter(s) in the config. 

> Hmm, I am not sure that I completely understood your point here, what do you mean by exceptions?

Given enough time, at some point, we'll need to break the back compatibility to keep the code manageable.

Training/python/2018v1/Training_v0p1.py
```diff
@@ -218,10 +204,21 @@ def compile_model(model, learning_rate):
...
+    gpus = tf.config.list_physical_devices('GPU')
+    if gpus:
+        try:
+            tf.config.set_visible_devices(gpus[data_loader.gpu_index], 'GPU')
+            tf.config.experimental.set_virtual_device_configuration(gpus[data_loader.gpu_index],
+                [tf.config.experimental.VirtualDeviceConfiguration(memory_limit=data_loader.gpu_mem*1024)])
+            logical_gpus = tf.config.experimental.list_logical_devices('GPU')
+            print(len(gpus), "Physical GPUs,", len(logical_gpus), "Logical GPUs")
+        except RuntimeError as e:
+            print(e)
```
kandrosov, line 218: Hm... it appears that these settings should be done before `TauLosses.SetSFs(*dataloader.TauLossesSFs)` and `create_model`, otherwise TensorFlow still allocates memory on other GPUs...

kandrosov (changes requested):

Training/python/2018v1/Training_v0p1.py
shedprog: Unfortunately, I did not test it on machines with several gpus

kandrosov (approved):

PR #59, opened 2021-08-31
Displaced Tau Reconstruction

shedprog:
This PR is created to see the difference (not for merge). 

The PR introduces the changes needed to run the reconstruction above PF-candidates level:
- modification of DataLoader
- modification pf the feature_scaling script
- modification of the python dataloader wrapper
- preliminary GNN architecture

### DIFF
--- a/Training/configs/conda_submit.job
+++ b/Training/configs/conda_submit.job
@@ -1,0 +1,6 @@
+Requirements = OpSysAndVer == "CentOS7"
+#Requirements = OpSysAndVer == "CentOS7" && (CUDAGlobalMemoryMb > 30000)
+Request_GPUs = 1
+RequestMemory = 30000
++RequestRuntime = 604800
+queue
--- a/Training/interface/DataLoaderReco_main.h
+++ b/Training/interface/DataLoaderReco_main.h
@@ -191,11 +191,9 @@ public:
             getVecRef(PfType::pfCand_numberOfPixelHits)    = getValue<Int_t>("nPixelHits",pref).at(idx_srt);
             getVecRef(PfType::pfCand_numberOfHits)         = getValue<Int_t>("nHits",pref).at(idx_srt);
             getVecRef(PfType::pfCand_hasTrackDetails)      = getValue<Int_t>("hasTrackDetails",pref).at(idx_srt);
-            getVecRef(PfType::pfCand_dxy)                  = getValue<Float_t>("dxy",pref).at(idx_srt);
             
             Float_t dz = getValue<Float_t>("dz",pref).at(idx_srt);
-            if(std::isnormal(dz) || dz == 0)
-              getVecRef(PfType::pfCand_dz) =  dz;
+
             getVecRef(PfType::pfCand_caloFraction)         = getValue<Float_t>("caloFraction",pref).at(idx_srt);
             getVecRef(PfType::pfCand_hcalFraction)         = getValue<Float_t>("hcalFraction",pref).at(idx_srt);
             getVecRef(PfType::pfCand_rawCaloFraction)      = getValue<Float_t>("rawCaloFraction",pref).at(idx_srt);
@@ -215,18 +213,27 @@ public:
             getVecRef(PfType::pfCand_E)                    = v.E();
             
             if(getValue<Int_t>("hasTrackDetails",pref).at(idx_srt))
-                getVecRef(PfType::pfCand_dxy_error)        = getValue<Float_t>("dxy_error", pref).at(idx_srt);
-                getVecRef(PfType::pfCand_dz_error)         = getValue<Float_t>("dz_error", pref).at(idx_srt);
-                getVecRef(PfType::pfCand_track_chi2)       = getValue<Float_t>("track_chi2", pref).at(idx_srt);
-                getVecRef(PfType::pfCand_track_ndof)       = getValue<Float_t>("track_ndof", pref).at(idx_srt);
-
-            Float_t jet_eta = tauTuple->get<Float_t>("jet_eta");
-            Float_t jet_phi = tauTuple->get<Float_t>("jet_phi");
-
-            getVecRef(PfType::jet_eta)        = jet_eta;
-            getVecRef(PfType::jet_phi)        = jet_phi;
-            getVecRef(PfType::pfCand_rel_eta) = getValue<Float_t>("eta",pref).at(idx_srt) - jet_eta;
-            getVecRef(PfType::pfCand_rel_phi) = DeltaPhi<Float_t>(getValue<Float_t>("phi",pref).at(idx_srt), jet_phi);
+            {
+                if(std::isnormal(dz) || dz == 0) 
+                    getVecRef(PfType::pfCand_dz) =  dz;
+                getVecRef(PfType::pfCand_dxy)        = getValue<Float_t>("dxy",pref).at(idx_srt);
+                if(std::isnormal(pfCand_dxy_error) || pfCand_dxy_error == 0)
+                    getVecRef(PfType::pfCand_dxy_error)  = getValue<Float_t>("dxy_error", pref).at(idx_srt);
+                if(std::isnormal(pfCand_dz_error) || pfCand_dz_error == 0)
+                    getVecRef(PfType::pfCand_dz_error)   = getValue<Float_t>("dz_error", pref).at(idx_srt);
+                getVecRef(PfType::pfCand_track_chi2) = getValue<Float_t>("track_chi2", pref).at(idx_srt);
+                getVecRef(PfType::pfCand_track_ndof) = getValue<Float_t>("track_ndof", pref).at(idx_srt);
+            }
+            
+            if(tauTuple->get<Int_t>("jet_index")>=0)
+            {
+                Float_t jet_eta = tauTuple->get<Float_t>("jet_eta");
+                Float_t jet_phi = tauTuple->get<Float_t>("jet_phi");
+                // getVecRef(PfType::jet_eta)        = jet_eta;
+                // getVecRef(PfType::jet_phi)        = jet_phi;
+                getVecRef(PfType::pfCand_deta) = getValue<Float_t>("eta",pref).at(idx_srt) - jet_eta;
+                getVecRef(PfType::pfCand_dphi) = DeltaPhi<Float_t>(getValue<Float_t>("phi",pref).at(idx_srt), jet_phi);
+            }
         }
       }
 
--- a/Training/interface/DataLoaderReco_main.h
+++ b/Training/interface/DataLoaderReco_main.h
@@ -98,6 +98,13 @@ public:
 
   private:
 
+      template <typename FeatureT>
+      const float Scale(const int idx, const float value, const bool inner)
+      {
+        return std::clamp((value - FeatureT::mean[idx][inner]) / FeatureT::std[idx][inner],
+                          FeatureT::lim_min[idx][inner], FeatureT::lim_max[idx][inner]);
+      }
+
       static constexpr float pi = boost::math::constants::pi<float>();
 
       template<typename Scalar>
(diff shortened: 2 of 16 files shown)


Training/interface/DataLoaderReco_main.h
```diff
@@ -0,0 +1,264 @@
...
+                size_t index = start_index + n_features * pfCand_i + _feature_idx;
+                data->x.at(index) = 
+                    Scale<typename  FeaturesHelper<decltype(_fe)>::scaler_type>(
+                        _feature_idx, value, false
+                        );
+            };
+
+            size_t idx_srt = indices.at(pfCand_i);
+            getVecRef(PfType::pfCand_valid                ,1.0);
+            getVecRef(PfType::pfCand_pt                   ,getValue<Float_t>("pt",pref).at(idx_srt));
```
kandrosov, line 191: NB. `getValue` by name is very slow... probably we should think about converting the dataset to `.hdf5` to improve the training speed. (no action required for this PR)

Training/interface/DataLoaderReco_main.h
```diff
@@ -0,0 +1,264 @@
...
+                    getVecRef(PfType::pfCand_dz_error  ,getValue<Float_t>("dz_error", pref).at(idx_srt));
+                if(std::isnormal(getValue<Float_t>("dxy_error", pref).at(idx_srt)))
+                    getVecRef(PfType::pfCand_dxy_error ,getValue<Float_t>("dxy_error", pref).at(idx_srt));
+
+                getVecRef(PfType::pfCand_dxy           ,getValue<Float_t>("dxy",pref).at(idx_srt));
+                getVecRef(PfType::pfCand_track_chi2    ,getValue<Float_t>("track_chi2", pref).at(idx_srt));
+                getVecRef(PfType::pfCand_track_ndof    ,getValue<Float_t>("track_ndof", pref).at(idx_srt));
+            }
+            
+            if(tauTuple->get<Int_t>("jet_index")>=0)
```
kandrosov, line 237: in the places where get can be avoided (like here) better to avoid using it :)

Training/interface/DataLoaderReco_main.h
```diff
@@ -0,0 +1,264 @@
...
+                getValue<Float_t>("phi",pref).at(idx_srt),
+                getValue<Float_t>("mass",pref).at(idx_srt)
+              );
+
+            getVecRef(PfType::pfCand_px, v.Px());
+            getVecRef(PfType::pfCand_py, v.Py());
+            getVecRef(PfType::pfCand_pz, v.Pz());
+            getVecRef(PfType::pfCand_E, v.E());
+            
+            if(getValue<Int_t>("hasTrackDetails",pref).at(idx_srt))
```
kandrosov, line 223: add `hasTrackDetails` as an input feature

Training/interface/DataLoader_main.h
```diff
@@ -762,7 +762,8 @@ class DataLoader {
             fillGrid(Br::muon_dphi, DeltaPhi(tau.muon_phi.at(idx), tau.tau_phi));
 
             fillGrid(Br::muon_dxy, tau.muon_dxy.at(idx));
-            fillGrid(Br::muon_dxy_sig, std::abs(tau.muon_dxy.at(idx)) / tau.muon_dxy_error.at(idx));
+            if(tau.muon_dxy_error.at(idx) != 0)
```
kandrosov, line 765: Do not modify `DataLoader_main.h` in this PR

Training/python/2018v1/Training_v0p1.py
```diff
@@ -204,7 +204,7 @@ def create_model(net_config):
 
 def compile_model(model, learning_rate):
     # opt = keras.optimizers.Adam(lr=learning_rate)
-    opt = tf.keras.optimizers.Nadam(learning_rate=learning_rate, beta_1=1e-4)
+    opt = tf.keras.optimizers.Nadam(learning_rate=learning_rate, schedule_decay=1e-4)
```
kandrosov, line 207: Do not modify `Training_v0p1.py` in this PR

Training/python/DataLoader.py
```diff
@@ -133,7 +131,7 @@ def compile_classes(file_config, file_scaling):
         _rootpath = os.path.abspath(os.path.dirname(__file__)+"/../../..")
         R.gROOT.ProcessLine(".include "+_rootpath)
 
-        _LOADPATH = "TauMLTools/Training/interface/DataLoader_main.h"
+        _LOADPATH = "TauMLTools/Training/interface/DataLoaderReco_main.h"
```
kandrosov, line 134: - read path to `.h` from `.yaml`
- Would be better to have a separate `DataLoader.py` for reco (one can extract the common logic into `DataLoaderBase` module and load it)

Training/configs/conda_submit.job
```diff
@@ -0,0 +1,6 @@
+Requirements = OpSysAndVer == "CentOS7"
```
kandrosov, line 1: Rename to `conda_submit_desy.job`

Training/python/feature_scaling.py
```diff
@@ -57,11 +61,13 @@
         raise ValueError('Specified file_range is not valid: should be either -1 (run on all files in file_path) or range [a, b] with a<=b')
     file_names_id = [fname.split('_')[-1].split('.root')[0] for fname in file_names] # id as taken from the name: used as file identifier (key field) in the output json files with quantiles
     #
-    dR_tau_outer_cone = cone_definition_dict['outer']['dR']
-    tau_pt_name, tau_eta_name, tau_phi_name = cone_selection_dict['TauFlat']['var_names']['pt'], cone_selection_dict['TauFlat']['var_names']['eta'], cone_selection_dict['TauFlat']['var_names']['phi']
-    inner_cone_min_pt = cone_definition_dict['inner']['min_pt']
-    inner_cone_min_radius = cone_definition_dict['inner']['min_radius']
-    inner_cone_opening_coef = cone_definition_dict['inner']['opening_coef']
+    # if cone_definition_dict != None:
```
kandrosov, line 64: feature_scaling should work for both reco and id. please add cones as an option and uncomment the code

Training/python/reco/Training_SNNv0.py
```diff
@@ -0,0 +1,367 @@
...
+    model.save("%s_final.tf" % train_name, save_format="tf")
+    return fit_hist
+
+
+config   = os.path.abspath( "../../configs/trainingReco_v1.yaml")
+scaling  = os.path.abspath("../../configs/scaling_params_vReco_v1.json")
+dataloader = DataLoader.DataLoader(config, scaling)
+input_map, input_shape, input_types  = dataloader.get_config()
+
+setup_main = {
```
kandrosov, line 347: move setup to `.yaml`

Training/python/reco/Training_SNNv0.py
```diff
@@ -0,0 +1,367 @@
...
+
+        x_em1 = self.embedding1(tf.abs(xx[:,:,self.map_features['pfCand_particleType']]))
+        x_em2 = self.embedding2(tf.abs(xx[:,:,self.map_features['pfCand_pvAssociationQuality']]))
+        x_em3 = self.embedding3(tf.abs(xx[:,:,self.map_features['pfCand_fromPV']]))
+        # x = self.normalize(xx)
+        # x = self.scale(x)
+
+        x_part1 = xx[:,:,:self.map_features['pfCand_particleType']]
+        x_part2 = xx[:,:,(self.map_features["pfCand_fromPV"]+1):]
+        x = tf.concat((x_em1,x_em2,x_em3,x_part1,x_part2),axis = 2)
```
kandrosov, line 187: better to have categorical features at the end of the tensor and have `input_dim`s for them as parameters in the constructor.

Training/python/reco/Training_SNNv0.py
```diff
@@ -0,0 +1,367 @@
...
+
+    fit_hist = model.fit(data_train, validation_data = data_val,
+                         epochs = data_loader.n_epochs, initial_epoch = data_loader.epoch,
+                         callbacks = callbacks)
+
+    model.save("%s_final.tf" % train_name, save_format="tf")
+    return fit_hist
+
+
+config   = os.path.abspath( "../../configs/trainingReco_v1.yaml")
```
kandrosov, line 342: pass config and output as command line parameters.

kandrosov (changes requested):

Training/interface/DataLoaderReco_main.h
shedprog: so, should I change from the reference `getValue<Float_t>("eta",pref).at(idx_srt)` to `tau.eta[idx_srt]` here and everywhere + for pfCand?

kandrosov: More like `tau.pfCand_eta[idx_srt]` etc. The idea of using prefix was to have the same code to read `lostTrack` and `pfCand` collections, but at the end also the scaling will be different, so I think it is more convenient to have explicit access to variables without using get & prefix.

Training/python/feature_scaling.py
shedprog: @kandrosov  This code is already compatible with both ID and RECO configs, these comments are more just to see the difference when reviewing. But may be a bit tricky to follow the logic, I used debugger for this. If needed, I can reshuffle code a bit, but I do not have ideas how to do it on-flight.

shedprog: About in particular these variables: they are somehow hidden inside scaling_utils.py(fill_aggregators) and addressed by demand

Training/python/reco/Training_SNNv0.py
shedprog: The problem with such generalization that I see, let's say you want to just cut the tail with this input_dim [:,:,:, -input_dim:], the embedding is still different for every of the categorical variables (they have different ranges), but with the current option, you can still adjust everything yourself in the Training_SNNv0.py code

kandrosov: Maybe it is better to have two separate input tensors: one with normal features other with categorical. What do you think?

shedprog: Hm, that would be ok I think, in this case we would avoid slicing op. and we can initiate input ranges for embendding from yaml, if we have `variable : [ null, categorical, a1, a2], a1 and at2 will not be concidered for scaling but we can use them for init for embendding

shedprog: @kandrosov there is one problem in the current NN: matrix with distances is generated on every call -> not efficient. This part should be taken out of recurrent call of GNN_layer https://github.com/shedprog/TauMLTools/blob/d50878fa6ec06a09933ac1e85f167df04f57c96a/Training/python/reco/Training_SNNv0.py#L60-L86

If I now create one more tensor, I have 2x calculations of distance matrices

kandrosov: you should append two tensors before passing them to GNN layer. i.e.
```
normal_inputs ------------>| 
cat_input1 -> emb_layer1 ->| -> concat -> GNN_layer1 -> ...
cat_input2 -> emb_layer2 ->|
```

shedprog: I see, I also understood that this transformation for evaluating if `w` is dependent on input `x` to MyGNNLayer because of the creation of `a` and `b` but I would think about alternative tensor operation not to recalculate `w` every time, the waigths de-facto are the same for all  `MyGNNLayer`

kandrosov: Yes, it could be an alternative approach. But the idea that the distances between points should change between layers is the core for the SNN architecture: we start with distances in (eta, phi) space and let the NN regrouping points for each consecutive layer in order to find a better parameter space for distance measurement.

Training/python/reco/Training_SNNv0.py
```diff
@@ -0,0 +1,394 @@
...
+    for checkpoint_dir in glob(f'{log_name}*.tf'):
+         mlflow.log_artifacts(checkpoint_dir, f"model_checkpoints/{checkpoint_dir}")
+    mlflow.log_artifacts(model_path, "model")
+    mlflow.log_artifacts(logs, "custom_tensorboard_logs")
+    mlflow.log_artifact(csv_log_file)
+
+    return fit_hist
+
+@hydra.main(config_path='.', config_name='train')
+def main(cfg: DictConfig) -> None:
```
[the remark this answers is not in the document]
yaourtpourtoi: Right, I also quite like how DataLoader classes inherit from each other with the current implementation. Then maybe once we converge on the stable version of `Training_v0p1.py` for the main DeepTau campaign, we could have a look at how the code can be restructured to account for usage of different models.

PR #60, opened 2021-09-14
MLflow integration into training

yaourtpourtoi:
This PR introduces the tracking of the training step with [MLflow](https://mlflow.org/docs/latest/index.html), configuration control with [hydra](https://hydra.cc/docs/intro/) and sets the scene for the upcoming changes in evaluation pipeline. Specifically:

* `DataLoader` interface has been modified: now instead of the path to the main training yaml cfg the corresponding python dictionary should be passed explicitly. The change is propagated to other scripts which make use of `DataLoader`, however #59 might require adaptation. 
* Input of args to the training `Training_v0p1.py` script is handled with hydra. The corresponding `train.yaml` "meta" cfg file is introduced to steer the parsing of arguments.
* MLflow logging of the training step. This includes both [auto](https://www.mlflow.org/docs/latest/tracking.html#automatic-logging) and manual logging of some parameters and artifacts (see corresponding lines with `mlflow.log_*` in `Training_v0p1.py`).
* Functionality to log additional files/parameters to MLflow either to existing run, or to a new one, see `log_to_mlflow.py` and accompanying `log_to_mlflow.yaml`.
* `tau-ml-env.yaml` has been updated with new packages.
*  Minor modifications such as moving of gpu arguments to `train.yaml` and splitting of `get_config()` into two separate methods (to avoid usage of global vars and to have more clarity).

The training step is to be run with: 
```bash
python Training_v0p1.py experiment_name=<user-specified-experiment-name>
```

This will run the training script with default parameter values as specified in the `train.yaml`. However, the user is very flexible in [overriding](https://hydra.cc/docs/advanced/override_grammar/basic) parameters from the command line. Note the usage of [default lists](https://hydra.cc/docs/tutorials/basic/your_first_app/defaults) in `train.yaml` to parse directly the training yaml file to the final configuration. This for example allows to override parameters from the training yaml file in the command line:

```bash
python Training_v0p1.py experiment_name=<user-specified-experiment-name> training_cfg.SetupNN.n_batches=10 training_cfg.Setup.n_tau=1
```

### DIFF
--- a/Training/python/2018v1/Training_v0p1.py
+++ b/Training/python/2018v1/Training_v0p1.py
@@ -137,7 +137,7 @@ def create_model(net_config):
             # n_comp_features = len(input_cell_external_branches) + len(net_config.comp_branches[comp_id])
             n_comp_features = len(net_config.comp_branches[comp_id])
             input_layer_comp = Input(name="input_{}_{}".format(loc, comp_name),
-                                     shape=(n_cells_eta[loc], n_cells_phi[loc], n_comp_features))
+                                     shape=(net_config.n_cells_eta[loc], net_config.n_cells_phi[loc], n_comp_features))
             input_layers.append(input_layer_comp)
             comp_net_setup.RecalcLayerSizes(n_comp_features, 2, 1)
             #input_layer_comp_masked = Masking(name="input_{}_{}_masking".format(loc, comp_name))(input_layer_comp)
@@ -155,7 +155,7 @@ def create_model(net_config):
         else:
             prev_layer = reduced_inputs[0]
         window_size = 3
-        current_size = n_cells_eta[loc]
+        current_size = net_config.n_cells_eta[loc]
         n = 1
         while current_size > 1:
             win_size = min(current_size, window_size)
@@ -176,13 +176,13 @@ def create_model(net_config):
         #dense_net_setup.RecalcLayerSizes(128, 1, 0.5, False)
         #final_dense = reduce_n_features_1d(features_concat, dense_net_setup, 'final')
         final_dense = dense_block_sequence(features_concat, dense_net_setup, 4, 'final')
-        output_layer = Dense(n_outputs, name="final_dense_last",
+        output_layer = Dense(net_config.n_outputs, name="final_dense_last",
                              kernel_initializer=dense_net_setup.kernel_init)(final_dense)
 
     else:
         final_dense = dense_block(features_concat, 1024, dense_net_setup,
                                   'tmp_{}'.format(net_config.name), 1)
-        output_layer = Dense(n_outputs, name="tmp_{}_dense_last".format(net_config.name),
+        output_layer = Dense(net_config.n_outputs, name="tmp_{}_dense_last".format(net_config.name),
                              kernel_initializer=dense_net_setup.kernel_init)(final_dense)
     softmax_output = Activation("softmax", name="main_output")(output_layer)
 
@@ -208,6 +208,7 @@ def run_training(train_suffix, model_name, model, data_loader, to_profile):
 
     gen_train = data_loader.get_generator(primary_set = True)
     gen_val = data_loader.get_generator(primary_set = False)
+    input_shape, input_types = data_loader.get_input_config()
 
     data_train = tf.data.Dataset.from_generator(
         gen_train, output_types = input_types, output_shapes = input_shape
@@ -243,18 +244,14 @@ with open(os.path.abspath( "../../configs/training_v1.yaml")) as f:
     config = yaml.safe_load(f)
 scaling  = os.path.abspath("../../configs/ShuffleMergeSpectral_trainingSamples-2_files_0_50.json")
 dataloader = DataLoader.DataLoader(config, scaling)
-netConf_full, input_shape, input_types  = dataloader.get_config()
-
-n_cells_eta = dataloader.n_cells
-n_cells_phi = dataloader.n_cells
-n_outputs = dataloader.tau_types
 
 setup_gpu(dataloader)
 
 TauLosses.SetSFs(*dataloader.TauLossesSFs)
 print("loss consts:",TauLosses.Le_sf, TauLosses.Lmu_sf, TauLosses.Ltau_sf, TauLosses.Ljet_sf)
-model = create_model(netConf_full)
 
+netConf_full = dataloader.get_net_config()
+model = create_model(netConf_full)
 compile_model(model, dataloader.learning_rate)
 fit_hist = run_training('step{}'.format(1), dataloader.model_name, model, dataloader, False)
 
--- a/Training/python/2018v1/Training_v0p1.py
+++ b/Training/python/2018v1/Training_v0p1.py
@@ -266,11 +266,10 @@ def main(cfg: DictConfig) -> None:
     
     # run the training with mlflow tracking
     with mlflow.start_run(**run_kwargs) as active_run:
+        setup_gpu(cfg.gpu_cfg)
         training_cfg = OmegaConf.to_object(cfg.training_cfg) # convert to python dictionary
         scaling_cfg = to_absolute_path(cfg.scaling_cfg)
         dataloader = DataLoader.DataLoader(training_cfg, scaling_cfg)
-
-        setup_gpu(dataloader)
 
         TauLosses.SetSFs(*dataloader.TauLossesSFs)
         print("loss consts:",TauLosses.Le_sf, TauLosses.Lmu_sf, TauLosses.Ltau_sf, TauLosses.Ljet_sf)
--- a/Training/python/2018v1/Training_v0p1.py
+++ b/Training/python/2018v1/Training_v0p1.py
@@ -215,7 +215,7 @@ def compile_model(model, learning_rate):
     metric_names = {(m if isinstance(m, str) else m.__name__): '' for m in metrics}
     mlflow.log_dict(metric_names, 'input_cfg/metric_names.json')
 
-def run_training(model, data_loader, to_profile, log_suffix, path_to_csv_log, path_to_log_dir, path_to_model):
+def run_training(model, data_loader, to_profile, log_suffix):
 
     gen_train = data_loader.get_generator(primary_set = True)
     gen_val = data_loader.get_generator(primary_set = False)
@@ -230,7 +230,7 @@ def run_training(model, data_loader, to_profile, log_suffix, path_to_csv_log, pa
 
     model_name = data_loader.model_name
     log_name = '%s_%s' % (model_name, log_suffix)
-    csv_log_file = f"{path_to_csv_log}/{log_name}.log"
+    csv_log_file = "metrics.log"
     if os.path.isfile(csv_log_file):
         close_file(csv_log_file)
         os.remove(csv_log_file)
@@ -238,7 +238,7 @@ def run_training(model, data_loader, to_profile, log_suffix, path_to_csv_log, pa
     time_checkpoint = TimeCheckpoint(12*60*60, log_name)
     callbacks = [time_checkpoint, csv_log]
 
-    logs = path_to_log_dir + '/' + log_name + '_' + datetime.now().strftime("%Y.%m.%d(%H:%M)")
+    logs = log_name + '_' + datetime.now().strftime("%Y.%m.%d(%H:%M)")
     tboard_callback = tf.keras.callbacks.TensorBoard(log_dir = logs,
                                                      profile_batch = ('100, 300' if to_profile else 0),
                                                      update_freq = ( 0 if data_loader.n_batches_log<=0 else data_loader.n_batches_log ))
@@ -248,7 +248,7 @@ def run_training(model, data_loader, to_profile, log_suffix, path_to_csv_log, pa
                          epochs = data_loader.n_epochs, initial_epoch = data_loader.epoch,
                          callbacks = callbacks)
 
-    model_path = f"{path_to_model}/{log_name}_final.tf"
+    model_path = f"{log_name}_final.tf"
     model.save(model_path, save_format="tf")
 
     # mlflow logs
@@ -284,8 +284,7 @@ def main(cfg: DictConfig) -> None:
         netConf_full = dataloader.get_net_config()
         model = create_model(netConf_full)
         compile_model(model, dataloader.learning_rate)
-        fit_hist = run_training(model, dataloader, False, cfg.log_suffix, 
-                                to_absolute_path(cfg.path_to_csv_log), to_absolute_path(cfg.path_to_log_dir), to_absolute_path(cfg.path_to_model))
+        fit_hist = run_training(model, dataloader, False, cfg.log_suffix)
 
         mlflow.log_dict(training_cfg, 'input_cfg/training_cfg.yaml')
         mlflow.log_artifact(scaling_cfg, 'input_cfg')
--- a/Training/python/2018v1/Training_v0p1.py
+++ b/Training/python/2018v1/Training_v0p1.py
@@ -1,4 +1,5 @@
 import os
+import yaml
 import gc
 import sys
 import glob
@@ -238,7 +239,8 @@ def run_training(train_suffix, model_name, model, data_loader, to_profile):
     return fit_hist
 
 
-config   = os.path.abspath( "../../configs/training_v1.yaml")
+with open(os.path.abspath( "../../configs/training_v1.yaml")) as f:
+    config = yaml.safe_load(f)
 scaling  = os.path.abspath("../../configs/ShuffleMergeSpectral_trainingSamples-2_files_0_50.json")
 dataloader = DataLoader.DataLoader(config, scaling)
 netConf_full, input_shape, input_types  = dataloader.get_config()
--- a/Training/python/2018v1/Training_v0p1.py
+++ b/Training/python/2018v1/Training_v0p1.py
@@ -21,6 +21,13 @@ from tensorflow.keras.layers import Input, Dense, Conv2D, Dropout, AlphaDropout,
 from tensorflow.keras.callbacks import Callback, ModelCheckpoint, CSVLogger
 from datetime import datetime
 
+import mlflow
+mlflow.tensorflow.autolog(log_models=False)
+
+import hydra
+from hydra.utils import to_absolute_path
+from omegaconf import DictConfig, OmegaConf
+
 sys.path.insert(0, "..")
 from common import *
 import DataLoader
@@ -204,7 +211,11 @@ def compile_model(model, learning_rate):
     ]
     model.compile(loss=TauLosses.tau_crossentropy_v2, optimizer=opt, metrics=metrics, weighted_metrics=metrics)
 
-def run_training(train_suffix, model_name, model, data_loader, to_profile):
+    # log metric names for passing them during model loading
+    metric_names = {(m if isinstance(m, str) else m.__name__): '' for m in metrics}
+    mlflow.log_dict(metric_names, 'input_cfg/metric_names.json')
+
+def run_training(model, data_loader, to_profile, log_suffix, path_to_csv_log, path_to_log_dir, path_to_model):
 
     gen_train = data_loader.get_generator(primary_set = True)
     gen_val = data_loader.get_generator(primary_set = False)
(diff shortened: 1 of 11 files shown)


tau-ml-env.yaml
```diff
@@ -8,8 +8,11 @@ dependencies:
   - absl-py=0.12.0=pyhd8ed1ab_0
   - afterimage=1.21=ha9998ff_1003
   - aiohttp=3.7.4=py38h497a2fe_0
+  - alembic=1.4.1=py_0
```
shedprog, line 11: This comment unrelated with the following PR but I have problem installing the environment from this yaml. When I do `conda env create -f tau-ml-env.yaml` I obtain the list of corresponding errors:

CondaVerificationError: The package for cudnn located at /nfs/dust/cms/user/mykytaua/softML/miniconda3/pkgs/cudnn-8.1.0.77-h90431f1_0
appears to be corrupted. The path 'include/cudnn.h'
specified in the package manifest cannot be found.

CondaVerificationError: The package for cudnn located at /nfs/dust/cms/user/mykytaua/softML/miniconda3/pkgs/cudnn-8.1.0.77-h90431f1_0
appears to be corrupted. The path 'include/cudnn_adv_infer.h'
specified in the package manifest cannot be found.
...

Training/python/2018v1/Training_v0p1.py
```diff
@@ -234,25 +248,44 @@ def run_training(train_suffix, model_name, model, data_loader, to_profile):
...
-setup_gpu(dataloader)
-
-TauLosses.SetSFs(*dataloader.TauLossesSFs)
-print("loss consts:",TauLosses.Le_sf, TauLosses.Lmu_sf, TauLosses.Ltau_sf, TauLosses.Ljet_sf)
-model = create_model(netConf_full)
-
-compile_model(model, dataloader.learning_rate)
-fit_hist = run_training('step{}'.format(1), dataloader.model_name, model, dataloader, False)
-
+@hydra.main(config_path='.', config_name='train')
```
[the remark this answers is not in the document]
yaourtpourtoi: If I understand it correctly, it can: `config_path` [specifies the relative location](https://hydra.cc/docs/advanced/search_path) of the yaml cfg with respect to the main script. However, I tried to run it from the different folder now and it doesn't work out-of-the-box because of the `sys.path.insert(0, "..")`. So this line should be changed accordingly to point to the directory with `common.py` and `DataLoader.py` modules.

tau-ml-env.yaml
yaourtpourtoi: Strange, I tested the fresh installation of the environment (`conda env create -n tau-ml-dev -f tau-ml-env.yaml`) and didn't see any errors... Do you still observe it if you try to install the environment from the yaml which is currently on the master branch (i.e. before me adding new packages)?

Training/python/DataLoader.py
```diff
@@ -322,27 +318,29 @@ def read_from_file(file_path):
...
-        input_tau_branches = get_branches(self.config,"TauFlat")
-        input_cell_pfCand_ele_branches = get_branches(self.config,"PfCand_electron")
-        input_cell_pfCand_muon_branches = get_branches(self.config,"PfCand_muon")
-        input_cell_pfCand_chHad_branches = get_branches(self.config,"PfCand_chHad")
-        input_cell_pfCand_nHad_branches = get_branches(self.config,"PfCand_nHad")
-        input_cell_pfCand_gamma_branches = get_branches(self.config,"PfCand_gamma")
-        input_cell_ele_branches = get_branches(self.config,"Electron")
-        input_cell_muon_branches = get_branches(self.config,"Muon")
+        input_tau_branches = DataLoader.get_branches(self.config,"TauFlat")
+        input_cell_pfCand_ele_branches = DataLoader.get_branches(self.config,"PfCand_electron")
```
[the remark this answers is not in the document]
yaourtpourtoi: good point, changed in 43d95f1

Training/python/2018v1/Training_v0p1.py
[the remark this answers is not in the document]
shedprog: If mlflow part is flexible in relation to the current directory, it is good, thanks. Yes, it should be changed to `sys.path.insert(0, os.path.dirname(__file__)+"/..")`

tau-ml-env.yaml
shedprog: For me, it does not work either

shedprog (approved):

kandrosov:
Thank you @yaourtpourtoi ! So far everything looks fine for me (both installation and running) except for two things:
- After running the code, I see 4 new folders + 1 log file (`DeepTau2018v0_step1_final.tf  DeepTau2018v0_step1.log  logs  mlruns  outputs`). Shouldn't everything be stored under `mlruns/exp_id/run_hash/...`)?
- I don't succeed to find `.yaml` with all params that would correspond to the running conditions in mlruns artefacts. Where can I find this information?

yaourtpourtoi:
> * After running the code, I see 4 new folders + 1 log file (`DeepTau2018v0_step1_final.tf  DeepTau2018v0_step1.log  logs  mlruns  outputs`). Shouldn't everything be stored under `mlruns/exp_id/run_hash/...`)?

Ah, right, I was a bit careless about the logs... So in 650cba9 I include model checkpoints and log with metrics, which I previously overlooked. Also I include custom tensorboard logs, however mlflow includes some by default (maybe they coincide) so this might be actually redundant. 

Then in 1f5ddad I remove path specification from input training configuration. Since everything is now in mlflow folders, I don't think it is needed for the user to specify the paths in the yaml cfg. In that case, during execution of the training script, everything will be stored into intermediate hydra log directory in `outputs/{data}/{time}` and from this directory then logged into mlflow run folders. Therefore `outputs` directory can be considered as a "backup" folder, and cleaned up if needed. 

> * I don't succeed to find `.yaml` with all params that would correspond to the running conditions in mlruns artefacts. Where can I find this information?

Do you mean the input configuration for `Training_v0p1.py` which is composed by hydra? If so, then I included this info as well in 4f314ce

tau-ml-env.yaml
shedprog: @yaourtpourtoi are you testing it on naf.desy or lxplus?

shedprog: ok, the issue it closed. It is working after cleaning of the conda base environment

shedprog:
I have doubts about folder naming, but I am not sure how flexible mlflow setups:
1) in the output folder, the directory tree by data/time/ is not so intuitive, if I have plenty of experiments running in parallel, then I don't know which experiment corresponds to which folder (I have to look inside_folder/inside_folder/inside_YAML file content). Maybe in this case, it would be easier to do something like <model_name>_output/<experiment_name> or to really anticipate overlapping (but do we really need this flaxibility?) something like <model_name>_output/<experiment_name>/<full_date>
2) in the folder mlflow, why we have two subfolders with names 0 and 1?
3) in the folder 2 we have the main content, does the names of the folder inside should be run_id-hash or it is flaxible, because if we have some flexibility in this we could do some naming which could be corresponded to the elements inside ./output/ (<model_name>_output/<experiment_name> or something...)

What do you think about it?

yaourtpourtoi:
> 1. in the output folder, the directory tree by data/time/ is not so intuitive, if I have plenty of experiments running in parallel, then I don't know which experiment corresponds to which folder (I have to look inside_folder/inside_folder/inside_YAML file content). Maybe in this case, it would be easier to do something like <model_name>_output/<experiment_name> or to really anticipate overlapping (but do we really need this flaxibility?) something like <model_name>_output/<experiment_name>/<full_date>

Surely, the names of hydra logs [can be changed](https://hydra.cc/docs/configure_hydra/workdir) and in general the [log behaviour](https://hydra.cc/docs/tutorials/basic/running_your_app/logging) too. However, I don't think that this is needed. The intended and user-friendly way to interact with mlflow is via [UI](https://mlflow.org/docs/latest/tracking.html#tracking-ui) and not via looking through the folders in `mlruns` or `outputs` (but this can of course be done). Executing `mlflow ui` in the directory with mlflow (the one which contains `mlruns` folder) will open a local port which can be opened in the browser to navigate through all experiments, times, names and files altogether (this is what I was showing during one of the meetings)

So the general logic is that the program stores everything in the corresponding `outputs/{data}/{time}` intermediate subdirectory (default hydra behaviour), then in the program the user logs what needs to be logged into mlflow and from that moment on interact only with mlruns/mlflow UI.  This way, logs which are created by hydra in `outputs` folder are left simply as a remnant / backup for those files which are already logged to mlflow.

> 2. in the folder mlflow, why we have two subfolders with names 0 and 1?

These are mlflow experiments with ID 0 and 1. Sometimes it happened to me that when the very first experiment is created / when the program is terminated with exception, a dummy experiment with corresponding ID appears with (correct me if I am wrong) just one file `meta.yaml`. I didn't really bother about it and the reason behind it because in the pipeline the referencing is done by experiment name, so user doesn't have to worry about which experiment ID should be provided to scripts (since these dummy experiments can break the order of "correct experiments"). By the way, experiment ID can be found also from mlflow UI main page.

> 3. in the folder 2 we have the main content, does the names of the folder inside should be run_id-hash or it is flaxible, because if we have some flexibility in this we could do some naming which could be corresponded to the elements inside ./output/ (<model_name>_output/<experiment_name> or something...)

No, unfortunately this naming convention can't be changed on the mlflow side. Moreover, if changed, it wouldn't allow for a file retrieval in the downstream evaluation pipeline. There, given the run ID as input to a script, sometimes I manually reference folders assuming that they are named as this run ID, so changing the naming pattern would cause ambiguity. However, there is a possibility to [name runs](https://stackoverflow.com/questions/57199472/is-it-possible-to-set-change-mlflow-run-name-after-run-initial-creation) to have better namings/search convenience within mlflow UI (btw it is basically an internal [tag](https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.set_tag) assignment). So I still do recommend to have mlflow UI opened all the time during running of experiments and inspect the runs from there. This workflow  should scale better with amount of runs comparing to manual inspection of folders inside `mlruns`.

kandrosov:
@yaourtpourtoi Thanks for the updates! Two follow up questions:
- Now everything seem to be available under mlruns directory. Do I understand correctly that after program is finished `outputs` directory can be safely removed?
- As you already answered above, the directory structure inside mlruns can't be changed, but is it possible to change how experiment id are generated? The sequential way is not working for a training distributed system. E.g. on my side, I'll have experiment "k_epx" with id 1, while Mykyta will have "m_exp" with the same id. This will prevent us from simply merging results into the central /eos directory. if it is possible to have experiment id as a hash (maybe shorter than run hash) it would solve the problem.

kandrosov:
It would also be good that exp_name -> exp_id conversion is deterministic. Like this, we can launch runs that belong to the same experiment on the different sites and after that merge them with a simple `rsync` command.

One other thing that must be fixed is that all paths in `mlruns` configs (except copies of the original config stored in artefacts) must be relative and not absolute. E.g. currently `artifact_location` in `meta.yaml` is absolute which is again will be site/user-dependent and will present us from merging results.

yaourtpourtoi:
> * Now everything seem to be available under mlruns directory. Do I understand correctly that after program is finished `outputs` directory can be safely removed without any loss of information?

Yes, from what I see after testing runs everything in hydra log folder is stored into mlflow.

> * As you already answered above, the directory structure inside mlruns can't be changed, but is it possible to change how experiment id are generated? The sequential way is not working for a training distributed system. E.g. on my side, I'll have experiment "k_exp" with id 1, while Mykyta will have "m_exp" with the same id. This will prevent us from simply merging results into the central /eos directory. if it is possible to have experiment id as a hash (maybe shorter than run hash) it would solve the problem.

Right, this is a good point, and after a quick check I am afraid that this is not possible... At least with available API, but I have a feeling that opening all `meta.yaml` files in all runs within the experiment and setting there experiment ID to a desired number should work. 

Another possibility which seems appealing to me is to look into [tracking server specification](https://www.mlflow.org/docs/latest/tracking.html#logging-to-a-tracking-server). The idea is to open a port on lxplus machine with eos and then listen to it on the machine where the training is running and tell mlflow to communicate directly with it. Or in any other way try to mount eos to the training machine and communicate with eos file system directly. Not sure that this is possible though, but if so, then mlflow will be able to automatically infer which experiments exist on eos and create new accordingly.

kandrosov:
I would prefer to avoid any network dependence
- opening ports on lxplus is not so easy
- many high performance computing sites have very strict rules for outside connections, so mounting eos or connecting to some external port could be not allowed. If we build our architecture around a central server it would limit our possibility of using such resources.

As current workaround I propose that we implement the following procedure:
1. create experiment in the central mlruns directory in /eos without runs
1. rsync from /eos to local site/usr
1. run trainings and sync back to central /eos

since run ids are unique, the procedure specified above should work (as far as I see). What do you think?
if you don't see hidden problems with that, the only thing that should be fixed before merging this PR is abs -> rel path in the mlruns yamls.

yaourtpourtoi:
> I would prefer to avoid any network dependence
> 
> * opening ports on lxplus is not so easy
> * many high performance computing sites have very strict rules for outside connections, so mounting eos or connecting to some external port could be not allowed. If we build our architecture around a central server it would limit our possibility of using such resources.

OK, I see, then indeed it makes sense to make the architecture more decentralised.

> As current workaround I propose that we implement the following procedure:
> 
> 1. create experiment in the central mlruns directory in /eos without runs
> 2. rsync from /eos to local site/usr
> 3. run trainings and sync back to central /eos
> 
> since run ids are unique, the procedure specified above should work (as far as I see). What do you think?
> if you don't see hidden problems with that, the only thing that should be fixed before merging this PR is abs -> rel path in the mlruns yamls.

Conceptually, this workflow looks good to me, however it would require some path adjustments. I did a bit of searching and found [this issue](https://github.com/mlflow/mlflow/issues/3144) which describes exactly our use case (and surprisingly includes also my suggestion with a bringing up a server and your arguments of why it doesn't work for us 😊 ) The problem as was mentioned is that the paths to run artifacts (specified in `meta.yaml` for each run) are determined at runtime and there is no interface to change that afterwards if the artefacts move around. This makes it not possible for `mlflow ui` to display them because it searches for artefacts in the `artifact_uri` and can't find it. On top of this we have a problem that experiment ID is not unique (oppositely to run ID), which makes combination of `mlruns` folders between those who run trainings not possible straight away (because probably they will all have experiment ID = 0 / 1 / 2).

So I would suggest to make a standalone script which given an mlflow experiment folder, recursively goes through it and resets the paths and experiment id/name in all `meta.yaml` files to the user-specified values. Basically, this can be considered as a preparation for moving to eos/`rsync`ing from another machine: the script is executed to change all the paths to eos folder (where the mlflow log are to be moved) and also setting the experiment id to the one which we agree to have in common (for rsyncing).  Would it make sense to go this way?

kandrosov:
@yaourtpourtoi how to modify parameters from the command line? I've tried several attempts like:
```sh
python Training_v0p1.py experiment_name=run3_cnn_ho1 '+SetupNN.conv_2d_setup=[ 3, 64, 2 ]'
```
or
```sh
python Training_v0p1.py experiment_name=run3_cnn_ho1 '+SetupNN={conv_2d_setup:[ 3, 64, 2 ]}
```
it runs but the actual values of the parameters are not changed.

shedprog:
> @shedprog could you, please, confirm that loss vs batch number plots (with step `n_batches_log`) for all the loss function components are recoverable from the information stored in mlflow run folder?

Technically - yes, but now the interface a bit confusing,

1) TB was interface friendly when everything was in one folder: (1) you can not just open all models together specifying path to the top folder -> it does not find all the tb-logs in subfolders, you have to specify with the comma explicitly path to every tensorboard file i.e: `--logdir <chosen_name_1>:outputs/2021-09-15/20-04-38/DeepTau2018v0_step1_2021.09.15\(20:05\)/train,--logdir <chosen_name_2>:outputs/2021-09-15/20-04-38/DeepTau2018v0_step1_2021.09.15\(20:05\)/val` (2) The folder name now is not informative it is hard to understand what you are comparing, unless you explicitly specify it as `<chosen_name_1>:`
image

2) Now in addition of per epoch_*, batch_* loss functions metrics I also have evaluation_* I am not sure what it means, maybe it is because I updated tb version, or mlfow also log something in TB logs.

3) @yaourtpourtoi I found the bug: if you do not finish the training, i.e: just run for one epoch, then models and logs are not copied to the mlflow, as usually we do not finish the training it may be painful

kandrosov:
> @yaourtpourtoi how to modify parameters from the command line? I've tried several attempts like:
> 
> ```shell
> python Training_v0p1.py experiment_name=run3_cnn_ho1 '+SetupNN.conv_2d_setup=[ 3, 64, 2 ]'
> ```
> 
> or
> 
> ```shell
> python Training_v0p1.py experiment_name=run3_cnn_ho1 '+SetupNN={conv_2d_setup:[ 3, 64, 2 ]}'
> ```
> 
> it runs but the actual values of the parameters are not changed.
> 
> P.S. same situation for any other parameters under `SetupNN` that I've tried

Please ignore the message above - it was my bad, I forgot to add "training_cfg.".

yaourtpourtoi:
> 3. @yaourtpourtoi I found the bug: if you do not finish the training, i.e: just run for one epoch, then models and logs are not copied to the mlflow, as usually we do not finish the training it may be painful

I am not sure that this is a bug because I can't do it differently. Since we don't have an explicitly defined training loop, I can't log the files to mlflow as they appear in the file system. Because I can't unfold the `fit()` method, it is possible to call `mlflow.log_*` only after `fit()` is done, that is after the training is finished. So in order to have more flexibility on mlflow logging, the `fit()` should be unfolded into explicit loop, where it would be possible to call mlflow logs e.g. after each epoch or n_batches. Or, have some fun (and it might be actually a good idea) and modify `mlflow` source code, because I found the other day that there are [some callbacks](https://github.com/mlflow/mlflow/blob/4d5d5c9e138013965335288aecd9ff956e1c4668/mlflow/tensorflow.py#L750) which are used internally in mlflow autologging and maybe we could use them explicitly.

yaourtpourtoi:
> 2. Now in addition of per epoch__, batch__ loss functions metrics I also have evaluation_* I am not sure what it means, maybe it is because I updated tb version, or mlfow also log something in TB logs.

Can it be the metrics on the validation set? As for the `custom_tensorboard_logs` folder, I didn't change anything in your implementation and just send what was there to mlflow. But as for the `tensorboard_logs`, these are mlflow automatic logs, and I would need to look at the source code of what is happening there

shedprog:
> > 1. @yaourtpourtoi I found the bug: if you do not finish the training, i.e: just run for one epoch, then models and logs are not copied to the mlflow, as usually we do not finish the training it may be painful
> 
> I am not sure that this is a bug because I can't do it differently. Since we don't have an explicitly defined training loop, I can't log the files to mlflow as they appear in the file system (at least within the scope of `Training_v0p1.py`). Because I can't unfold the `fit()` method, it is possible to call `mlflow.log_*` only after `fit()` is done, that is after the training is finished.
> 
> So in order to have more flexibility on mlflow logging, the `fit()` should be unfolded into explicit loop, where it would be possible to call mlflow logs e.g. after each epoch or n_batches. Or, have some fun (and it might be actually a good idea) and modify mlflow source code, because I found the other day that there are [some callbacks](https://github.com/mlflow/mlflow/blob/4d5d5c9e138013965335288aecd9ff956e1c4668/mlflow/tensorflow.py#L750) which are used internally in mlflow autologging and maybe we could use them explicitly in the training code.

Ok, I think for our hyperparameter tuning it still should be fine. However, I don't see why do you want to modify `fit()` all logging in our case is done by demand of callbacks, you can add to this callback to do some mlflow.log() or some copy to keep the synchronization between folders or anything else

yaourtpourtoi:
> Ok, I think for our hyperparameter tuning it still should be fine. However, I don't see why do you want to modify `fit()` all logging in our case is done by demand of callbacks, you can add to this callback to do some mlflow.log() or some copy to keep the synchronization between folders or anything else

Ah, you are right, my bad that I overlooked the option of logging inside of a `TimeCheckpoint` callback... I can fix it in this PR if needed

yaourtpourtoi:
As discussed above, there might be conflicts when we start merging each others experiment together under common eos area. In 2789982b I wrote a script which helps those unification by resetting all the paths and experiment IDs to those specified by the user. 

An exemplary workflow might be the following. I have 3 experiments with ID = {0, 2, 3} in `$MYDIR/Training/python/2018v1/mlruns` and for example, someone might have experiments with ID = {0, 1} in `$WORKDIR/Training/python/2018v1/mlruns`. With this setting, in meta.yaml files for each run there will be `artifact_uri: file://$MYDIR/Training/python/2018v1/mlruns/{experiment_ID}/{run_ID}/artifacts`. Also, in the main experiment `meta.yaml` there will be `artifact_location: file://$MYDIR/Training/python/2018v1/mlruns/{experiment_ID}`. In other person's configs instead of $MYDIR there will be $WORKDIR.

Then, suppose I want to move my experiment 1 to eos and merge it with runs from other's experiment 0.  If we simply copy all the folders into `$EOSDIR/mlruns`, join run subfolders in folder 1 with those of folder 0 under common folder, say, 10, running `mlflow ui` from $EOSDIR will not show any artifacts. 
1. Firstly, because `artifact_location` needs to be set to `$EOSDIR/mlruns/10` and the same change needs to be propagated to `artifact_uri` _for each run_.
2. Secondly, once we merged run subfolder of experiment 0 with experiment 1 together into experiment 10, this new common experiment ID hasn't been propagated into all the yaml files (into corresponding key `experiment_id`). 

So `set_mlflow_paths.py` does exactly that: loops over input experiment folder and resets all `meta.yaml` paths and experiment ids to what is needed. Then, the exemplary use case above can be approach as follows (variations possible):
1. Experiment folders 0 and 1 are moved to $EOSDIR/mlruns/0 and $EOSDIR/mlruns/1
2. Run subfolders from e.g. experiment 1 are moved to experiment 0. `1/meta.yaml` shouldn't be moved
3. Assuming  `set_mlflow_paths.py` is in $EOSDIR, execute the following command. This will rename folder 0 to 10, set all `experiment_id` to 10, reset the paths in yaml files so that they point to $EOSDIR/mlruns (instead $WORKDIR and $MYDIR), set `experiment_name` to "common_name":
```bash
python set_mlflow_paths.py -p ./mlruns -id 0 -np $EOSDIR/mlruns -nid 10 -nn common_name
```

4. Execute `mlflow ui` (NB: from $EOSDIR, that is from the directory which contains mlruns folder ) to bring up the server with UI

kandrosov (approved):

kandrosov:
Thank you @yaourtpourtoi ! The core functionality to use mlflow is implemented and working. I'm going to merge this PR. Additional features and improvements can be done in future PRs.

PR #67, opened 2021-10-15
Evaluation pipeline [part 1]

yaourtpourtoi:
This PR introduces the first step of the evaluation pipeline, namely `apply_training.py` script to run inference. Conceptually, it works as follows: model is fetched based on the given mlflow runID , applied to a given input file and then file with predictions is stored to artifacts of the given runID. The scripts's input arguments are specified in corresponding `apply_training.yaml` cfg file and they are:

* `training_cfg_upd`: a dictionary those parameters are then merged into/override the default training configuration taken from `path_to_training_cfg`. At this PR several parameters are reset in order to configure `DataLoader` to be used for inference. The key once are `n_tau=1` (DataLoader will yield 1 tau/batch), `include_mismatched=True` (see below), `validation_split=0.` (fraction of input dataset to set aside for validation). 
* `path_to_mlflow`: path to `mlruns` folder where the `experiment_id` will be searched
* `experiment_id`: ID of the mlflow experiment to which the given `run_id` belongs to
* `run_id`: mlflow runID which contains the info for a training of interest
* `path_to_training_cfg`: path to a training `*.yaml` configuration file. Default value is set to be the one logged to artifacts
* `scaling_cfg`: path to a `*.json` file with scaling parameters. Default value is set to be the one currently used in the training
* `path_to_file`: path to input file for which predictions are to be produced
* `file_alias`: unique string to identify this file. Used as a part of prediction file name, it will be needed downstream to refer to these predictions. 

An example of running would be:
```sh
python apply_training.py path_to_mlflow=../Training/python/2018v1/mlruns experiment_id=0 run_id=f790a2ad286c4d72ac549c8d902108a1 path_to_file=eval_data/TTToHadronic.root file_alias=TTToHadronic
```

Next, c27a5ed adds a switch `include_mismatched` to include gen mismatched entries. This is needed since now `DataLoader` [drops](https://github.com/cms-tau-pog/TauMLTools/blob/45babb742f1a15f96950e84d60c95acc9b65f7e9/Training/interface/DataLoader_main.h#L243) such events from yielding, and therefore number of entries in the input file is not equal to the number of predicted entries. This breaks the downstream evaluation code which assumes these numbers to be equal, therefore for the purpose of inference this flag is reset to `True` (default=False for training).

Lastly, git commit hash is now also logged to mlflow during the training (see 0ebbb76). This is needed mostly to keep track of the `DataLoader` source code which was used in the given training. Since `apply_training.py` by default imports the `DataLoader` class which is present at the runtime repo state, it might cause conflicts with the `DataLoader` class actually used during the training. 

At the moment, this step isn't automated, so if the current git repo state diverged from repo state used in the training, **user need to manually checkout that commit** and run `apply_training.py` from there.

### DIFF
--- a/Evaluation/apply_training.py
+++ b/Evaluation/apply_training.py
@@ -22,7 +22,10 @@ def main(cfg: DictConfig) -> None:
     # set up paths & gpu
     mlflow.set_tracking_uri(f"file://{to_absolute_path(cfg.path_to_mlflow)}")
     path_to_artifacts = to_absolute_path(f'{cfg.path_to_mlflow}/{cfg.experiment_id}/{cfg.run_id}/artifacts/')
-    setup_gpu(cfg.gpu_cfg)
+    if cfg.gpu_cfg is not None:
+        setup_gpu(cfg.gpu_cfg)
+    else:
+        os.environ["CUDA_VISIBLE_DEVICES"]="-1"
 
     # load the model
     with open(to_absolute_path(f'{path_to_artifacts}/input_cfg/metric_names.json')) as f:
@@ -36,18 +39,18 @@ def main(cfg: DictConfig) -> None:
         training_cfg = OmegaConf.merge(training_cfg, cfg.training_cfg_upd)
     training_cfg = OmegaConf.to_object(training_cfg)
 
-    # fetch historic git commit used to run training 
-    with mlflow.start_run(experiment_id=cfg.experiment_id, run_id=cfg.run_id) as active_run:
-        train_git_commit = active_run.data.params['git_commit'] if 'git_commit' in active_run.data.params else None
+    if cfg.checkout_train_repo: # fetch historic git commit used to run training
+        with mlflow.start_run(experiment_id=cfg.experiment_id, run_id=cfg.run_id) as active_run:
+            train_git_commit = active_run.data.params.get('git_commit')
 
-    # stash local changes and checkout 
-    if train_git_commit is not None:
-        repo = git.Repo(to_absolute_path('.'), search_parent_directories=True)
-        if cfg.verbose: print(f'\n--> Stashing local changes and checking out training commit: {train_git_commit}\n')
-        repo.git.stash('save')
-        repo.git.checkout(train_git_commit)
-    else:
-        if cfg.verbose: print('\n--> Didn\'t find git commit hash in run artifacts, continuing with current repo state\n')
+        # stash local changes and checkout 
+        if train_git_commit is not None:
+            repo = git.Repo(to_absolute_path('.'), search_parent_directories=True)
+            if cfg.verbose: print(f'\n--> Stashing local changes and checking out training commit: {train_git_commit}\n')
+            repo.git.stash('save', 'stored_stash')
+            repo.git.checkout(train_git_commit)
+        else:
+            if cfg.verbose: print('\n--> Didn\'t find git commit hash in run artifacts, continuing with current repo state\n')
 
     # instantiate DataLoader and get generator
     import DataLoader
@@ -100,8 +103,8 @@ if __name__ == '__main__':
     except Exception as e:
         print(e)
     finally:
-        print(f'\n--> Checking out back branch: {current_git_branch}\n')
-        repo.git.checkout(current_git_branch)
-        if repo.git.stash('list') != '':
+        if 'stored_stash' in repo.git.stash('list'):
+            print(f'\n--> Checking out back branch: {current_git_branch}')
+            repo.git.checkout(current_git_branch)
             print(f'--> Popping stashed changes\n')
             repo.git.stash('pop')
--- a/Evaluation/apply_training.py
+++ b/Evaluation/apply_training.py
@@ -42,7 +42,7 @@ def main(cfg: DictConfig) -> None:
        
     # time_checkpoints = [time.time()]
     input_file_name = to_absolute_path(cfg.path_to_file)
-    output_file_name = cfg.file_alias
+    output_file_name = cfg.file_alias + '_pred'
     with uproot.open(input_file_name) as f:
         t = f['taus']
         n_taus = len(t['evt'].array())
@@ -68,7 +68,7 @@ def main(cfg: DictConfig) -> None:
 
     # store into intermediate hdf5 file
     predictions = pd.DataFrame({f'node_{tau_type}': predictions[:, int(idx)] for idx, tau_type in tau_types_names.items()})
-    targets = pd.DataFrame({f'target_{tau_type}': targets[:, int(idx)] for idx, tau_type in tau_types_names.items()}, dtype=np.int64)
+    targets = pd.DataFrame({f'node_{tau_type}': targets[:, int(idx)] for idx, tau_type in tau_types_names.items()}, dtype=np.int64)
     predictions.to_hdf(f'{output_file_name}.h5', key='predictions', mode='w', format='fixed', complevel=1, complib='zlib')
     targets.to_hdf(f'{output_file_name}.h5', key='targets', mode='r+', format='fixed', complevel=1, complib='zlib')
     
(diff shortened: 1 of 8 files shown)


Evaluation/apply_training.py
```diff
@@ -0,0 +1,81 @@
...
+    training_cfg = OmegaConf.load(to_absolute_path(cfg.path_to_training_cfg))
+    if cfg.training_cfg_upd is not None:
+        training_cfg = OmegaConf.merge(training_cfg, cfg.training_cfg_upd)
+    training_cfg = OmegaConf.to_object(training_cfg)
+
+    # instantiate DataLoader and get generator
+    scaling_cfg  = to_absolute_path(cfg.scaling_cfg)
+    dataloader = DataLoader.DataLoader(training_cfg, scaling_cfg)
+    gen_predict = dataloader.get_predict_generator()
+    tau_types_names = training_cfg['Setup']['tau_types_names']
```
shedprog, line 41: I think the problem that you described before related to the fact that `${path_to_mlflow}/${experiment_id}/${run_id}/artifacts/input_cfg/training_cfg.yaml` requires few more variables, in this case `self.dataloader_core = config["Setup"]["dataloader_core"]` that is not present in the initial training_v1.yaml. I think it can be solved by simple adding of the variable in the training artifacts/input_cfg. On the other hand, we will always modify data loader, and the freshest version of the Evaluation should use the freshest version of the data loader, for this we can 1) take into account it when modifying data loader 2) you can just merge the configs with the priority of config in artifacts. That would  add missing variables but also avoid changes in setups.

Training/interface/DataLoader_main.h
```diff
@@ -254,6 +254,14 @@ class DataLoader {
...
               ++tau_i;
             }
+            else {
+              if (include_mismatched)
+                ++tau_i;
+            }
+          }
+          else {
+            if (include_mismatched)
+              ++tau_i;
```
shedprog, line 264: Maybe, as previously suggested by Konstantin, we can try to do the following:
1) Return n_batches on every iteration (not just n_tau=1) 
2) add "running" n_batches_current variable in `Data class` so that when you do reshape here: https://github.com/cms-tau-pog/TauMLTools/blob/45babb742f1a15f96950e84d60c95acc9b65f7e9/Training/python/DataLoader.py#L160 you reshape by this n_batches_current to be consistent on the last iteration when tau_i < n_batches.

What @kandrosov think about it?

Evaluation/apply_training.py
yaourtpourtoi: Exactly, adding more variables/files to artifact cfg files isn't a maintable approach as we evolve with DataLoader development. So what I think should work is within the eval script do `git checkout <commit-hash>` where `commit-hash` is taken from artifacts and it corresponds to the commit used to run the training. This seems to me like a more general approach to synchronise training and evaluation steps: during training we log training/runtime specific files (fetched as artifacts at eval step) and the rest (structural modules/files) is commited to git (synchronised by `git checkout` at eval step).

However, we need to agree to run the training with the code to be in synch with git (which kinda makes sense)

Training/interface/DataLoader_main.h
yaourtpourtoi: Hm, you mean basically to reshape at all the steps by `self.batch_size` except for the last iteration, where the reshape is by `self.n_total_tau - self. n_batches_current*self. batch_size `?

Evaluation/apply_training.py
```diff
@@ -0,0 +1,81 @@
...
+    output_file_name = cfg.file_alias + '_pred'
+    with uproot.open(input_file_name) as f:
+        t = f['taus']
+        n_taus = len(t['evt'].array())
+    
+    predictions = []
+    targets = []
+    print(f'\n\n--> Processing file {input_file_name}, number of taus: {n_taus}')
+    for X,y in tqdm(gen_predict(input_file_name), total=n_taus/training_cfg['Setup']['n_tau']):
+        predictions.append(model.predict(X))
```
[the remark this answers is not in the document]
yaourtpourtoi: good point! added it in 7373718

Evaluation/apply_training.py
yaourtpourtoi: ok, addressed it in 48c9fe2

Evaluation/apply_training.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # load baseline training cfg and update it with parsed arguments
+    training_cfg = OmegaConf.load(to_absolute_path(cfg.path_to_training_cfg))
+    if cfg.training_cfg_upd is not None:
+        training_cfg = OmegaConf.merge(training_cfg, cfg.training_cfg_upd)
+    training_cfg = OmegaConf.to_object(training_cfg)
+
+    # fetch historic git commit used to run training 
+    with mlflow.start_run(experiment_id=cfg.experiment_id, run_id=cfg.run_id) as active_run:
+        train_git_commit = active_run.data.params['git_commit'] if 'git_commit' in active_run.data.params else None
```
kandrosov, line 41: - Please, add the possibility to disable the checkout step (keeping the current behaviour by default). In some cases, one could want to have a modified DataLoader. 
- if `params` is `dict`-like use `get` instead

Evaluation/apply_training.py
```diff
@@ -0,0 +1,107 @@
...
+
+sys.path.insert(0, "../Training/python")
+from common import setup_gpu
+
+@hydra.main(config_path='.', config_name='apply_training')
+def main(cfg: DictConfig) -> None:
+    # set up paths & gpu
+    mlflow.set_tracking_uri(f"file://{to_absolute_path(cfg.path_to_mlflow)}")
+    path_to_artifacts = to_absolute_path(f'{cfg.path_to_mlflow}/{cfg.experiment_id}/{cfg.run_id}/artifacts/')
+    setup_gpu(cfg.gpu_cfg)
```
kandrosov, line 25: What parameters should be set to run inference on CPU-only?

Training/interface/DataLoader_main.h
kandrosov: For the first version, it is fine. Improvements for computing performance can be done later, once we have the whole chain functional.
However, please avoid have `if(include_mismatched)` in two places. Add it at the end, e.g.:
```c++
if(!tau_is_set && include_mismatched)
  ++tau_i;
```

kandrosov (changes requested):
Only minor comments. Thank you @yaourtpourtoi !

Evaluation/apply_training.py
yaourtpourtoi: Sure, added `checkout_train_repo` switch in 06ae65f. As for `params`, this is an attribute of the mlflow class, so can't change it to a dict-like interface here

Training/interface/DataLoader_main.h
yaourtpourtoi: done! 06ae65f

Evaluation/apply_training.py
yaourtpourtoi: yes, now it can be done by setting `gpu_cfg: null` in the yaml config file

Evaluation/apply_training.py
kandrosov: sorry, is it possible or not to use `get`? Because in the current version of the code it is changed to `get`, but based on your reply it shouldn't be done.

yaourtpourtoi: Well, since it is a nested call, it is not possible for the first steps ([`data`](https://www.mlflow.org/docs/latest/python_api/mlflow.entities.html#mlflow.entities.Run.data) and [`params`](https://www.mlflow.org/docs/latest/python_api/mlflow.entities.html#mlflow.entities.RunData.params) are the class properties) but possible for `git_commit` (key in the dictionary) 🙂 I changed the behaviour for `git_commit` to `get()` because in that case it will directly return `None` if this key doesn't exist. So I don't have to explicitly check this with `if 'git_commit' in active_run.data.params` and it looks more compact

kandrosov (approved):

PR #69, opened 2021-11-01
Some updates and small fixes to the Shuffle&Merge scripts

DennRoy:
In ShuffleMergeSpectral:
- Added option to include all remaining Taus in the last job, which are otherwise neglected because the total number of Taus didn't evenly divide between all jobs
- Minor verbose fix
- Fix for lastbin_takeall: Scale=1 if bin is empty
- Added verbose to show the number of remaining Taus to be filled, as well as the status after one sample group has finished (showing the number of Taus that are neglected)

In the law tasks.py:
- Added new implemented option from above, and lastbin-takeall which was missing previously

In validation_tool:
- Added option to save output in other filetypes (default is unchanged, just pdf)
- Titles of output include the name of the sample type, Tau type or group name (instead of just indices)
- Also prints the groups belong to the numbered indices for the "uh_dataset_group_id" plot

### DIFF
--- a/Analysis/bin/ShuffleMergeSpectral.cxx
+++ b/Analysis/bin/ShuffleMergeSpectral.cxx
@@ -20,7 +20,7 @@
 
 namespace analysis {
 
-void sumBasedSplit(const std::vector<size_t>& files_entries, const size_t job_idx, const size_t n_job,
+void sumBasedSplit(const std::vector<size_t>& files_entries, const size_t job_idx, const size_t n_job, const bool overflowjob,
                   std::pair<size_t, size_t>& point_entry,  std::pair<size_t, size_t>& point_exit, size_t& step)
 { 
   // sumBasedSplit splits files into n_job sub-intervals
@@ -42,6 +42,10 @@ void sumBasedSplit(const std::vector<size_t>& files_entries, const size_t job_id
    // .second - index of last event
   point_entry = find_idx(job_idx, false);
   point_exit = find_idx(job_idx, true);
+  if(overflowjob && job_idx+1 == n_job){
+    step = step + (std::accumulate(files_entries.begin(), files_entries.end(), 0) % n_job);
+    point_exit = std::make_pair(files_entries.size()-1, files_entries[files_entries.size()-1]-1);
+  }
 };
 
 enum class MergeMode { MergeAll = 1 };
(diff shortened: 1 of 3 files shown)


Analysis/bin/ShuffleMergeSpectral.cxx
```diff
@@ -398,6 +404,7 @@ class SpectrumHists {
           double dis_scale = std::min(1.0, exp_disbalance*
                              ttype_entries.at(type)->Integral(0,binNx,binNy-1,binNy)*
                              ttype_prob.at(type)->GetBinContent(x,y));
+          if(dis_scale<=0.0) dis_scale = 1.0;
```
shedprog, line 407: I think this remark indeed makes sense, I agree, however I am looking at it from the following point of view: it can occur only in case when ttype_entries.at(type)->Integral(0,binNx,binNy-1,binNy) == 0, in this case our required disbalance is simply unreachable, so maybe we should raise an exception? In the opposite case, user will have what he did not expected from his input arguments.

kandrosov: I agree, better to throw an exception, because integral=0 is an unexpected input.

Production/scripts/validation_tool.py
```diff
@@ -32,9 +33,10 @@
 ROOT.gROOT.SetBatch(not args.visual)
 ROOT.gStyle.SetOptStat(0)
 
-pdf_dir = '/'.join([args.output, 'pdf'])
-if not os.path.exists(pdf_dir):
-  os.makedirs(pdf_dir)
+for formats in args.formats.split(","):
+  pdf_dir = '/'.join([args.output, formats])
```
kandrosov, line 37: better to use `os.path.join`

Production/scripts/validation_tool.py
```diff
@@ -44,10 +46,10 @@
 
 ## binning of tested variables (dataset group id and dataset id are guessed from jsons)
 BINS = {
-  'tau_pt'    : (50, 0, 5000),
-  'tau_eta'   : (5, -3.2, 3.2),
-  'tauType' : (10, 0, 10),
-  'sampleType': (20, -1, 19),
+  'tau_pt'    : (100, 0, 1000),
+  'tau_eta'   : (10, -2.4, 2.4),
```
kandrosov, line 50: keep +/-2.5 by default

kandrosov (changes requested):

Analysis/bin/ShuffleMergeSpectral.cxx
DennRoy: Thanks for your comments! I made the changes you asked for.

PR #70, opened 2021-11-03
Minor scaling update + documentation

yaourtpourtoi:
In this PR:
* minor clean-ups/fixes are made to scaling computation scripts + added `mask_nan` functionality
* outdated json files and redundant notebook are removed
* plotting scripts for scaling validation cleaned up + added flexibility in the choice of var_types
* README updated to describe feature scaling computation and validation procedure


Training/python/scaling_utils.py
```diff
@@ -237,43 +237,53 @@ def mask_inf(var_array, var_name=None, var_inf_counter=None):
...
         var_array = ak.mask(var_array, is_inf_mask, valid_when=False)
         if var_inf_counter is not None:
             var_inf_counter[var_name].append(np.sum(is_inf_mask) / ak.count(var_array))
     return var_array
 
-# def fill_aggregators(var_array, tau_eta_array, tau_phi_array, constituent_eta_array, constituent_phi_array,
-#                      var, var_type, file_i, file_name_id, cone_type, dR_tau_signal_cone, dR_tau_outer_cone,
-#                      sums, sums2, counts, fill_scaling_params=False, scaling_params=None, quantile_params=None):
-def fill_aggregators(tree, var, var_type, file_i, file_name_id, cone_type, cone_definition_dict, cone_selection_dict, inf_counter,
+def mask_nan(var_array, var_name=None, var_nan_counter=None):
```
kandrosov, line 246: Why do we need this method? Having NaN that was not filtered by pre-condition will fire back during the training, so early we spot this error, the better it is. Maybe an exception should be raised instead?

kandrosov (changes requested):

Training/python/scaling_utils.py
yaourtpourtoi: That's a good point and was also thinking about how to do this. So in the end I decided to stick to an approach of computing scaling params anyway but keep track of appearing NaNs and printing them in the end:

https://github.com/yaourtpourtoi/TauMLTools/blob/52de583a5e64cc3d4102eea51b98bb999099848f/Training/python/feature_scaling.py#L153-L154

It was okay for the current workflow which looked like: scaling params are computed with dropped NaNs and print shows that there are NaNs in feature X -> check is added to DataLoader for this feature -> scaling params don't have to be recomputed because they are already in synch with the DataLoader.

However I agree, this print might befor example  overlooked, so I will add a flag to `mask_nan()/mask_inf()` to raise an exception.

kandrosov (approved):

PR #71, opened 2021-11-05
Update feature scaling code to not scale disabled features

aloeliger:
I ran into some issues with the feature scaling code attempting to create scalings for things that were in the `Features_disable` portion of `training_v1.yaml`, which crashed because these features were disabled because they were empty.

I have added a quick check to disable feature scaling for things that are already disabled in the training config, which works for my purposes, to prevent it scaling empty features. I figured I would check whether this would be useful in the main branch.

### DIFF
--- a/Training/python/feature_scaling.py
+++ b/Training/python/feature_scaling.py
@@ -28,6 +28,7 @@ if __name__ == '__main__':
     # read cfg parameters
     setup_dict = scaling_dict['Scaling_setup']
     features_dict = scaling_dict['Features_all']
+    disabledFeatures = [variableName for sublist in (scaling_dict['Features_disable'][featureType] for featureType in scaling_dict['Features_disable']) for variableName in sublist] # get the list of disabled features, and flatten it.
     #
     assert type(args.var_types) == list
     if args.var_types[0] == "-1" and len(args.var_types) == 1:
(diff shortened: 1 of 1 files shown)


kandrosov:
The "disabled features" mechanism is meant to be used only at the training stage to simplify R&D on the NN inputs. If some features are not well defined (like, for example, `pfCand_dxy` for PF photons), they should be removed from the `.yaml` config.

yaourtpourtoi:
Also, in the proposed update scaling params of disabled features will still be present in the output json file. This is because their is [`init_dictionaries()`](https://github.com/cms-tau-pog/TauMLTools/blob/89e5ddbfaa481464cbfc66d763b5b63697055c40/Training/python/scaling_utils.py#L64) function called here:
https://github.com/cms-tau-pog/TauMLTools/blob/89e5ddbfaa481464cbfc66d763b5b63697055c40/Training/python/feature_scaling.py#L62

It will initialise mean/std (differently depending on their scaling type) for all the features present in `features_dict`, i.e. all the features as in .yaml cfg including "disabled" ones. Therefore despite they are skipped from the computation, their inititalised values will still be in the output json which is sort of confusing. So as Konstantin mentioned it might be a better approach to remove problematic features from the yaml cfg if they are not meant to be used in the training anyway.

aloeliger:
I believe I was getting index and attribute errors simply removing them from the configuration, but perhaps I was mistaken. If this is not useful I will close the PR.

PR #72, opened 2021-11-11
Evaluation pipeline [part 2]

yaourtpourtoi:
This is a PR which introduces the core part of the evaluation pipeline. Below is its schematic sketch and more details about its design concepts can be found in these presentations: [[1]](https://indico.cern.ch/event/1066000/contributions/4482034/attachments/2293019/3898990/DeepTau_eval.pdf), [[2]](https://indico.cern.ch/event/1067541/contributions/4489002/attachments/2295081/3903198/DeepTau_eval_1.pdf), [[3]](https://indico.cern.ch/event/1068999/contributions/4495448/attachments/2297277/3907278/DeepTau_eval_2.pdf)

image

<details>
  <summary>Click for a short intro</summary>

After the training is completed, the model receives a unique mlflow `run_id` (see [`Training_v0p1.py`](https://github.com/cms-tau-pog/TauMLTools/blob/master/Training/python/2018v1/Training_v0p1.py)) which is used later on to refer to the model and to assign to it & access files (aka _artifacts_). Model predictions are such an example, see [`apply_training.py`](https://github.com/cms-tau-pog/TauMLTools/blob/master/Evaluation/apply_training.py) where these `*.h5` files are logged to a provided mlflow `run_id`.  

Once the predictions are in place, one can evaluate the model performance in various regions across the phase space. This is done with `Evaluation/evaluate_performance.py` with input configuration steered by hydra. The main yaml cfg is `configs/run3.yaml` and therein it imports other configurations from `configs` folder (see `defaults` key). The idea is the following: given the model (as mlflow `run_id`), the corresponding to it `discriminator/*.yaml` configuration and original input + predictions + targets files for genuine taus and `vs_type`, dump for a given phase space region (`pt_bins`, `vs_type`, `sample_tau`, etc) the metrics to `performance.json` file. This json is also stored in the mlflow ecosystem under the corresponding`run_id` and so far includes the ROC curve data compatible with `RocCurve` class defined in `eval_tools.py` and `plot_roc.py` (so far, duplicated). 

Based on this json file, one can produce ROC curves with `Evaluation/plot_roc.py` for specified discriminators and specified phase space region (pt bins/vs_type/sample). Again, all the communication to retrieve necessary files is done based on mlflow run IDs, see the corresponding `Evaluation/plot_roc.yaml` file.

</details>

Taking as an example a model with `experiment_name=run3_cnn_ho2`, `experiment_id=2`, `run_id=06f9305d6e0b478a88af8ea234bcec20` stored currently on eos, ROC curves (e.g. "vs jet") with comparison to a standard DeepTau v2  can be produced with the following workflow. Note that demo original files and corresponding predictions were copied to `predictions` folder of this run ID beforehand.

Firstly, create an empty run ID for DeepTau v2 model. Note that you can also skip this step and use `RUN_ID=1161f0b0d0ca4ad290eac1c28b4eeb2b` which I already created 
```shell
python Training/python/log_to_mlflow.py path_to_mlflow=/eos/cms/store/group/phys_tau/TauML/prod_2018_v2/mlruns experiment_name=run3_cnn_ho2 files_to_log=null params_to_log=null
``` 

This will create a new `RUN_ID` which will be referred to later on. Now, evaluate performance vs jet for the current training with:
```shell
export INPUT_DIR=/afs/cern.ch/work/o/ofilatov/Tau/TauMLTools/Evaluation/eval_data # this is the path to original input ROOT files used for producing demo predictions
export EOS_DIR=/eos/cms/store/group/phys_tau/TauML/prod_2018_v2/mlruns
python Evaluation/evaluate_performance.py path_to_mlflow=$EOS_DIR discriminator=DeepTau_run3 experiment_id=2 run_id=06f9305d6e0b478a88af8ea234bcec20 vs_type=jet sample_tau=GluGluHToTauTau_M125 sample_vs_type=TTToSemiLeptonic path_to_input_taus=$INPUT_DIR/GluGluHToTauTau_M125.root path_to_input_vs_type=$INPUT_DIR/TTToSemiLeptonic.root
```

This will create (or update) and log to a corresponding mlflow run ID `performance.json` file with computed  ROC curve for the specified input configuration (see `RocCurve` class definition/`performance.json` for more details on the format used internally). Now doing the same but for the DeepTau v2p1:
```shell
export RUN_ID= # insert here the run id created above
python Evaluation/evaluate_performance.py path_to_mlflow=$EOS_DIR discriminator=DeepTau_v2p1 experiment_id=2 run_id=$RUN_ID vs_type=jet sample_tau=GluGluHToTauTau_M125 sample_vs_type=TTToSemiLeptonic path_to_input_taus=$INPUT_DIR/GluGluHToTauTau_M125.root path_to_input_vs_type=$INPUT_DIR/TTToSemiLeptonic.root path_to_pred_taus=null path_to_pred_vs_type=null path_to_target_taus=$EOS_DIR/2/06f9305d6e0b478a88af8ea234bcec20/artifacts/predictions/GluGluHToTauTau_M125_pred.h5 path_to_target_vs_type=$EOS_DIR/2/06f9305d6e0b478a88af8ea234bcec20/artifacts/predictions/TTToSemiLeptonic_pred.h5
```
Note that now the path to targets is specified separately and predictions are set to `null` which will read them from original input files.

Having these ROC curves computed, they can be plotted against each other with `plot_roc.py` script. Firstly, update `Evaluation/plot_roc.yaml` cfg to look like:
```yaml
path_to_mlflow: /eos/cms/store/group/phys_tau/TauML/prod_2018_v2/mlruns
experiment_id: 2
discriminators: 
  06f9305d6e0b478a88af8ea234bcec20: ['roc_curve']
  $RUN_ID: ['roc_curve'] # put here RUN_ID for DeepTau v2p1
reference: 
  $RUN_ID: 'roc_curve' # put here RUN_ID for DeepTau v2p1

# region
vs_type: jet 
sample_tau: GluGluHToTauTau_M125
sample_vs_type: TTToSemiLeptonic
pt_bin: [20,100] 

# will log the plot to the run of the first discriminator in the "discriminators" list
output_name: 'roc_curve_vs_${vs_type}_pt_${pt_bin[0]}_${pt_bin[1]}'
```
and run the script:

```shell
python Evaluation/plot_roc.py pt_bin=[100,1000] # it is always possible to override arguments from CL
```

Resulted pdf will be stored under `run_id=06f9305d6e...` artifacts and can be viewed directly or through mlflow UI. 
Lastly, in the similar fashion one can produce ROC curves for the other `vs_type` (e/mu) with a dedicated `DYJetsToLL_M-50-amcatnloFXFX_ext2` demo sample. 

<details>
  <summary>Known minor issues/potential improvements (to be done in the following PRs)</summary>

  * Make single `RocCurve` and `PlotSetup` classes (now duplicated in `plot_roc.py` and `eval_tools.py`)
  * ~~More elegant management of WPs~~
  * Improve and generalise selection cutflow
  * Original input root files aren't mlflow traceable
  * Implement weight options into eval workflow (so far it simply inherits the previous approach w/o adaptation)
  * Implement model cross-reference: store for each run information, in which runs this model appears in comparison plots
</details>


Evaluation/evaluate_performance.py
```diff
@@ -0,0 +1,174 @@
...
+    if discriminator.from_tuple:
+        read_branches.append(discriminator.pred_column)
+        if discriminator.wp_column != discriminator.pred_column:
+            read_branches.append(discriminator.wp_column)
+
+    # read original data with corresponging predictions into DataFrame
+    if path_to_input_vs_type is None:
+        df_all = eval_tools.create_df(path_to_input_taus, read_branches, path_to_pred_taus, path_to_target_taus, cfg.discriminator.pred_column_prefix, path_to_weights_taus)
+        # apply gen. selection
+        inclusive_gen_selection = ' or '.join([f'(gen_{tau_type}==1)' for tau_type in ['tau', cfg.vs_type]]) # gen_* are constructed in `add_targets()`
```
kandrosov, line 52: For example, if I want to evaluate roc for "taus from H->tautau" vs "jets from DY", at which point it is selected that taus should be taken from "H->tautau" and jets from "DY"? At this point, two datasets are already merged, so such selection should happen before, but then why it is also needed here?

yaourtpourtoi: Actually, the behaviour you described is the one assumed by default, but maybe it is not articulated clearly in the code. The "if" case which you referenced above I inherited from the previous code, and I think the logic behind it was to cover the case when both taus and jets are taken from the same single source. 

So setting `path_to_input_vs_type=None` triggers this case and both taus and jets are taken from `path_to_input_taus`, with predictions taken from `path_to_pred_taus ` and targets from `path_to_target_taus `. Then, one can apply this inclusive OR selection in order to select both types.

On the contrary, if one wants to have taus from HTT and jets from DY, `path_to_input_vs_type` should be specified to DY file and then it triggers the "else" case, which takes taus from `path_to_input_taus`, other type from `path_to_input_vs_type` and applies gen matching separately:
```python
    else:
        df_taus = eval_tools.create_df(path_to_input_taus, read_branches, path_to_pred_taus, path_to_target_taus, cfg.discriminator.pred_column_prefix, path_to_weights_taus)
        df_vs_type = eval_tools.create_df(path_to_input_vs_type, read_branches, path_to_pred_vs_type, path_to_target_vs_type, cfg.discriminator.pred_column_prefix, path_to_weights_vs_type)
        df_taus = df_taus.query('gen_tau==1')
        df_vs_type = df_vs_type.query(f'gen_{cfg.vs_type}==1')
        df_all = df_taus.append(df_vs_type)
```

I was working on it for quite a long time so I got used to this logic, but not sure now that it is the most intuitive one:)

kandrosov: I see. So it is some merge of new and old approaches... I think it would be preferable that we have two yaml configs one to run prediction other to run plots and then have tau type filters specified there. In this way, in the code, if/else wouldn't be necessary.
E.g. in the prediction yaml for DY we want to run predictions for all types (e,mu,tau,jet). While for rocs one can want to select (tau from HTT vs jet from DY) or (tau from DY vs jet from DY), or some crazy scenarios like (a mix of taus from HTT and DY vs a mix of jets from W and TT) etc. I.e. in the roc config, we don't need to define separate inputs "taus" and "other", but rather have a set of entries (file_name_pattern, tau_selection to apply to it).

yaourtpourtoi: > I.e. in the roc config, we don't need to define separate inputs "taus" and "other", but rather have a set of entries (file_name_pattern, tau_selection to apply to it).

@kandrosov Well, I think this logic can't be implemented at the `plot_roc.py` step, but can be (and in fact is already) included at `evaluate_performance.py` step.  In general, the idea of `plot_roc.py` is that it just extracts an entry from `performance.json` based on `sample` labels (where `tau` and `vs_type` come from) and plots them, so all the selections/roc computation need to be performed at the earlier stage.

But as I mentioned various scenarios can be realised and dumped into `performance.json` already at `evaluate_performance.py` step. `apply_training.py` requires just _some_ input ROOT file to produce predictions, so it can be e.g. `DY=jets+mu+ele` mixture, or already some preselected/mixture-of-samples file. 

Since it is `evaluate_performance.py` where ROC is computed, it will _always_ require two inputs (labelled as `*_taus` and `*_vs_type`) - but there is a freedom here to pick those inputs. The only assumption which is made is that on one will be applied the selection `gen_tau==1`, on the other `gen_{vs_type}==1` and a common cut on their combination. Then the ROC is computed and is stored in `performance.json` with identifiers `sample_tau/sample_vs_type/vs_type/pt_bin`. Based on this identifiers, it is then retrieved and plotted by `plot_roc.py`.

That means, that one can pass the same DY input file to `*_taus` and `*_vs_type` arguments (input/predictions/targets) to take both taus and jets from it. Or, inputs can be DY for taus and HTT for jets. As for the mixture case (e.g. W+TT for jets, HTT+DY for taus), this input needs to be constructed manually beforehand (e.g. `hadd`-ed) and be presented as two tuples of _single_ files with (h5 with predictions/ h5 with targets/ h5 with weights/ ROOT with inputs) one per tau/vs_type to `evaluate_performance.py` with a corresponding label `sample_vs_type=W_TT`, `sample_tau=HTT_DY`. Then in `plot_roc` one can plot this ROC curve by providing those labels.

Hopefully, I managed to explain the idea more or less clearly. So finally, coming back to the original "if/else" discussion, I think "if" case can be simply removed, leaving the "else" (two single inputs) case left. The variability of scenarios can then be included as I described above. Would this logic make sense?

yaourtpourtoi:
Restructured the workflow for plotting WPs (subjectively) to have better clarity of cfg arguments. As a test, this code produces the following ROC curve + WPs for:
```python
run_id='06f9305d6e0b478a88af8ea234bcec20' # DeepTau_run3
vs_type='jet'
sample_tau='GluGluHToTauTau_M125'
sample_vs_type='TTToSemiLeptonic'
```

image

Evaluation/evaluate_performance.py
kandrosov: thanks for the detailed reply @yaourtpourtoi ! I probably still didn't understand the treatment of `gen_tau` and `vs_type` datasets should be done differently. I don't particularly like the idea of using `hadd`. Wouldn't it be more ergonomic to define samples in `run3.yaml` in this way: 
```yaml
samples:
  DY: [ 'tau' ]
  TT: [ 'jet' ]
  W: [ 'tau', 'jet' ]
  ...
path_to_input: 'eval_data/${sample}*.root'
path_to_pred: '${path_to_mlflow}/${experiment_id}/${run_id}/artifacts/predictions/${sample}*.h5'
...
```
and then in `evaluate_performance.py`: simply loop over all files an apply corresponding selection. In this way, also multiple files for the single sample are supported without requiring special treatement.

yaourtpourtoi: Ah, I see, this is indeed an elegant way to treat multiple inputs - will add it to this PR

yaourtpourtoi:
Updated the way input samples are combined in `evaluate_performance.py` with the comments above + added a documentation for the evaluation pipeline.

kandrosov (approved):

kandrosov:
> Updated the way input samples are combined in `evaluate_performance.py` with the comments above + added a documentation for the evaluation pipeline.

thank you @yaourtpourtoi. All looks good.

PR #76, opened 2021-12-04
A few changes

DennRoy:
- Training/interface/DataLoader_main.h: Saves histograms of Taus over the rebinned pt/eta plane. If any bin is empty, the training crashes, so it's useful for debugging, and to see where any high imbalance may be.
- Training/python/feature_scaling.py: Small verbose change
- Training/python/plot_quantile_ranges.py & Training/python/plot_scaling_convergence.py: Avoid opening plots, as the scripts are stalled as long as a plot is displayed (at least, this is what happens for me)
- Training/python/scaling_utils.py: Treatment in feature scaling if a variable has no entries (or very few, like only 1). This is how i handled missing "pfCand_gamma" track information, instead of commenting it out.

### DIFF
--- a/Training/interface/DataLoader_main.h
+++ b/Training/interface/DataLoader_main.h
@@ -194,6 +194,7 @@ public:
             ("w_1_"+tau_name).c_str(),
             ("w_1_"+tau_name).c_str()
         ));
+        hist_weights[tau_type]->SaveAs(("Temp_"+tau_name+".root").c_str()); // It's required that all bins are filled in these histograms; save them to check incase binning is too fine and some bins are empty
 
         target_histogram.reset();
         input_histogram .reset();
--- a/Training/interface/DataLoader_main.h
+++ b/Training/interface/DataLoader_main.h
@@ -194,7 +194,7 @@ public:
             ("w_1_"+tau_name).c_str(),
             ("w_1_"+tau_name).c_str()
         ));
-        hist_weights[tau_type]->SaveAs(("Temp_"+tau_name+".root").c_str()); // It's required that all bins are filled in these histograms; save them to check incase binning is too fine and some bins are empty
+        if (debug) hist_weights[tau_type]->SaveAs(("Temp_"+tau_name+".root").c_str()); // It's required that all bins are filled in these histograms; save them to check incase binning is too fine and some bins are empty
 
         target_histogram.reset();
         input_histogram .reset();
--- a/Training/python/plot_quantile_ranges.py
+++ b/Training/python/plot_quantile_ranges.py
@@ -118,9 +118,11 @@ def main(
 
                 ### fetch clamping params
                 mean = var_scaling['mean']
-                clamp_range_left = mean + var_scaling['lim_min']*var_scaling['std']
-                clamp_range_right = mean + var_scaling['lim_max']*var_scaling['std']
-                clamp_range = [clamp_range_left, clamp_range_right]
+                has_None = mean==None
+                if not has_None:
+                    clamp_range_left = mean + var_scaling['lim_min']*var_scaling['std']
+                    clamp_range_right = mean + var_scaling['lim_max']*var_scaling['std']
+                    clamp_range = [clamp_range_left, clamp_range_right]
 
                 ### fetch quantiles
                 median = var_quantiles['median']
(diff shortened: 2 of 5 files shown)


Training/python/scaling_utils.py
```diff
@@ -173,11 +173,11 @@ def compute_mean(sums, counts, aggregate=True, *file_range):
...
         if file_range:
             assert len(file_range) == 2 and file_range[0] <= file_range[1]
-            return sums[file_range[0]:file_range[1]].sum()/counts[file_range[0]:file_range[1]].sum()
+            return sums[file_range[0]:file_range[1]].sum()/counts[file_range[0]:file_range[1]].sum() if counts[file_range[0]:file_range[1]].sum()>1 else 0.
         else:
-            return sums.sum()/counts.sum()
+            return sums.sum()/counts.sum() if counts.sum()>1 else 0.
     else:
-        return sums/counts
+        return sums/counts if counts>1 else 0.
```
yaourtpourtoi, line 180: As a minor thing, since `counts` is an array, having it in if statement won't return a single bool (but an array of bool), so I think here (and similarly in `compute_std()`) there should be `counts.sum()>1` too (or `any(counts)>1`):
```suggestion
        return sums/counts if counts.sum()>1 else 0.
```

Training/python/scaling_utils.py
```diff
@@ -173,11 +173,11 @@ def compute_mean(sums, counts, aggregate=True, *file_range):
     if aggregate:
         if file_range:
             assert len(file_range) == 2 and file_range[0] <= file_range[1]
-            return sums[file_range[0]:file_range[1]].sum()/counts[file_range[0]:file_range[1]].sum()
+            return sums[file_range[0]:file_range[1]].sum()/counts[file_range[0]:file_range[1]].sum() if counts[file_range[0]:file_range[1]].sum()>1 else 0.
         else:
-            return sums.sum()/counts.sum()
+            return sums.sum()/counts.sum() if counts.sum()>1 else 0.
```
yaourtpourtoi, line 178: I am thinking that having 0(1) for mean(std) as return values in case of few counts might be misleading for further plotting and mean/std analysis. Would it be more suitable to return `None` and have a printout like "low statistics in {var} for mean/std computation" in that case? 

Of course having `mean=0` and `std=1` values stored in scaling json is OK in the sense that effectively it won't apply scaling to it. But in `DataLoader` there is also [clamping](https://github.com/cms-tau-pog/TauMLTools/blob/d233f566dfd733fa01cdf7986feb816882555b09/Training/interface/DataLoader_main.h#L320), therefore in that case feature by default will be clamped to interval [-5, 5], which might not be entirely correct. So having it as a `None` would be a sign that some manual treatment is needed.

Training/python/scaling_utils.py
```diff
@@ -216,6 +216,7 @@ def get_quantiles(var_array):
     """
     quantile_dict = {}
     var_array = ak.to_numpy(ak.flatten(var_array, axis=-1))
+    if np.size(var_array)==0: return {'median': 0., 'min': 0., 'max': 0., '1sigma': {'left': 0., 'right': 0.}, '2sigma': {'left': 0., 'right': 0.}, '3sigma': {'left': 0., 'right': 0.}, '5sigma': {'left': 0., 'right': 0.}}
```
yaourtpourtoi, line 219: The same comment here as for `compute_mean()/compute_std()` functions, maybe it's better to return `None` too?

yaourtpourtoi (commented):
@DennRoy thanks for adding missing cases and updating the scaling part! I left a few comments inline in the code for discussion.

Training/python/scaling_utils.py
DennRoy: Ok, I changed this, as well as in the other parts, but having `None` values in the quantiles breaks plot_quantile_ranges.py, so I added a special treatment there too

Training/python/plot_quantile_ranges.py
```diff
@@ -145,16 +147,18 @@ def main(
...
-                suspicious_dict['left_within'] = clamp_range_left > two_sigma_left
-                suspicious_dict['right_within'] = clamp_range_right < two_sigma_right
-                # suspicious_dict['left_beyond'] = clamp_range_left < five_sigma_left
-                # suspicious_dict['right_beyond'] = clamp_range_right > five_sigma_right
-                suspicious_dict['one_sigma_empty'] = one_sigma_left == one_sigma_right
-                suspicious_dict['two_sigma_empty'] = two_sigma_left == two_sigma_right
-                suspicious_dict['three_sigma_empty'] = three_sigma_left == three_sigma_right
-                suspicious_dict['five_sigma_empty'] = five_sigma_left == five_sigma_right
+                has_None = has_None or (None in [median, min_value, max_value, one_sigma_left, two_sigma_left, three_sigma_left, five_sigma_left, one_sigma_right, two_sigma_right, three_sigma_right, five_sigma_right])
+                if not has_None:
```
yaourtpourtoi, line 151: So given possible `None` presence this can be changed into:

```python
suspicious_dict = {}
is_suspicious = False
has_None = None in [mean, median, min_value, max_value, one_sigma_left, two_sigma_left, three_sigma_left, five_sigma_left, one_sigma_right, two_sigma_right, three_sigma_right, five_sigma_right])
if not has_None:
    clamp_range_left = mean + var_scaling['lim_min']*var_scaling['std']
    clamp_range_right = mean + var_scaling['lim_max']*var_scaling['std']
    clamp_range = [clamp_range_left, clamp_range_right]
    # suspicious_dict['...'] = ... # all the suspicious definitions as before
    is_suspicious = any(suspicious_dict.values())
elif mean is None:
    print(f'       {var_name}, {cone_type}: mean is empty')
    continue
else:
    print(f'       {var_name}, {cone_type}: quantiles are empty')
    continue   
``` 
Then, the remaining part starting from `if is_suspicious or not only_suspicious:` doesn't require any changes w.r.t. current master branch.

yaourtpourtoi:
Thanks a lot @DennRoy! All looks good to me in PR.

Training/interface/DataLoader_main.h
```diff
@@ -194,6 +194,7 @@ class DataLoader {
             ("w_1_"+tau_name).c_str(),
             ("w_1_"+tau_name).c_str()
         ));
+        hist_weights[tau_type]->SaveAs(("Temp_"+tau_name+".root").c_str()); // It's required that all bins are filled in these histograms; save them to check incase binning is too fine and some bins are empty
```
kandrosov, line 197: the histogram should be saved only when debug mode is on (e.g. add `debug` parameter in .yaml config)

kandrosov (changes requested):

Training/interface/DataLoader_main.h
DennRoy: Ok, I fixed this (didn't push yet), but I'm wondering: Is there a way to check if a `debug` parameter is initialized in the yaml config? Right now I have an additional line to check for this in Trainingv0p1.py and apply_training.py (would it be needed anywhere else?), but I was hoping there's some way to add this check in DataLoader_main.h.

kandrosov: if you define a variable in `Setup` section of `training.yaml` it should be accessible in c++ (like, for example `n_tau` etc.)

DennRoy: Yes, but my question was, what if it's not defined? It would cause a crash for everyone working with their current local configurations if they don't add that variable.

kandrosov: oh, they will need to add it, otherwise, the code will crash. However, the change is minimal, so I think it is ok to apply it now (there is not that many people who actively run training and the patch for their configs will be trivial).

yaourtpourtoi: I also think that this isn't a problem to add `debug` option. In general, each training is bound to use the version of DataLoader with which it was run. That means that every training is always in conflict with all future developments of DataLoader class. Basically, the main conflict point for current developments is `apply_training` when the DataLoader is used to load the data into the model. So to handle divergence of developments, there is a checkout of git commit used to run that given training:
   https://github.com/cms-tau-pog/TauMLTools/blob/d233f566dfd733fa01cdf7986feb816882555b09/Evaluation/apply_training.py#L51

In that perspective, your change won't affect the inference step of existing trainings, because in any case they use the past version of DataLoader (or in my case even more complicated workarounds, so your changes won't complicate them further at all :) )

kandrosov (approved):

PR #84, opened 2022-02-17
L1 reg, input selection, tauscore cut and weight options

lucasrussell01:
Added: 
- Missing debug option in config
- Optional first layer regularisation for tau dense and convolutional layers
- Optional DeepTau VSjet cut for data selection
- Option to toggle weights on/off
- Option to disable entire feature groups (eg TauFlat)
- Option to disable inner or outer cone
- Option to exclude area of inner cone from outer cone

### DIFF
--- a/Training/interface/DataLoader_main.h
+++ b/Training/interface/DataLoader_main.h
@@ -242,7 +242,7 @@ public:
           const auto sample_type = static_cast<analysis::SampleType>(tau.sampleType);
           bool tau_is_set = false;
 
-          if (gen_match &&tau.tau_byDeepTau2017v2p1VSjetraw >DeepTauvsVSjet_cut){
+          if (gen_match &&tau.tau_byDeepTau2017v2p1VSjetraw >DeepTauVSjet_cut){
             if (recompute_tautype){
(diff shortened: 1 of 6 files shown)


Training/interface/DataLoader_main.h
```diff
@@ -869,7 +869,11 @@ class DataLoader {
                   const bool inside_iso_cone = dR < iso_cone;
                   if(inner && !inside_signal_cone) continue;
                   // if(!inner && (inside_signal_cone || !inside_iso_cone)) continue;
-                  if(!inner && !inside_iso_cone) continue;
+                  if(rm_inner_from_outer){
+                    if(!inner && (!inside_iso_cone || inside_signal_cone)) continue;
+                  } else {
+                    if(!inner && !inside_iso_cone) continue;
+                  }
```
kandrosov, line 876:
```c++
const bool accept_inner = inner && inside_signal_cone;
const bool accept_outer = !inner && inside_iso_cone && (!rm_inner_from_outer || !inside_signal_cone);
if(!(accept_inner || accept_outer)) continue;
```

kandrosov (changes requested):

PR #85, opened 2022-02-20
Updated the core of the DataLoader

shedprog:
This fix in the core is aiming to modify two aspects:

Issue with inconsistent number of batches in different epochs:
- [x] Allow to load tensors with not fully filled taus in LoadData() function
- [x] Crop input tensor correspondingly to this number in the DataLoader base
- [x] Modify QueueEx and DataSource classes in order to collect the rests on the last iteration and repack them by n taus * n times + rests * 1 time (to allow use all the taus available)
- [ ] Propagate changes to the Reco study

Unimaginably large loading time for get_predict_generator:
- [x] Allow get_predict_generator load by n_taus for n_taus >> 1 + a bit less on last iteration (with this fix loading one file [GluGluHToTauTau_M125/eventTuple_1-10.root] is taking by apply_training.py in **~10min**)
- [x] Allow masking of taus with undefined type when model.predict(X) in apply_training.py - will bring few more times speed up for the MC when we have a lot of not matched to TauType candidates.

### DIFF
--- a/Training/python/2018v1/train.yaml
+++ b/Training/python/2018v1/train.yaml
@@ -10,7 +10,7 @@ path_to_mlflow: mlruns
 experiment_name: ???
 
 # setup
-scaling_cfg: ../../configs/ShuffleMergeSpectral_trainingSamples-2_files_0_50.json # for DataLoader initialisation
+scaling_cfg: ../../configs/ShuffleMergeSpectral_trainingSamples-2_rerun_files_0_19.json # for DataLoader initialisation
 gpu_cfg:
   gpu_mem  : 7 # in Gb
   gpu_index: 0
--- a/Training/python/2018v1/train.yaml
+++ b/Training/python/2018v1/train.yaml
@@ -10,7 +10,7 @@ path_to_mlflow: mlruns
 experiment_name: ???
 
 # setup
-scaling_cfg: ../../configs/ShuffleMergeSpectral_trainingSamples-2_rerun_files_0_19.json # for DataLoader initialisation
+scaling_cfg: ../../configs/ShuffleMergeSpectral_trainingSamples-2_files_0_498.json # for DataLoader initialisation
 gpu_cfg:
   gpu_mem  : 7 # in Gb
   gpu_index: 0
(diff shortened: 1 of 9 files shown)


shedprog:
@yaourtpourtoi can you remind please the following: do we need to predict something for taus when they y_onehot vector is simply = [0, 0, 0, 0] it is just ignored in the future, can I just mask them and set prediction = [0, 0, 0, 0]? When they are ignored?

yaourtpourtoi:
Yes, indeed that's a good idea! These entries are implicitly discarded in `evaluate_performance.py` anyway, so we don't have to run inference on them:
https://github.com/cms-tau-pog/TauMLTools/blob/75d39fed20a628a964bfc8a2be80df1dc24efc14/Evaluation/evaluate_performance.py#L53-L54

shedprog:
@kandrosov This PR can be merged

Training/python/2018v1/train.yaml
```diff
@@ -10,7 +10,7 @@ path_to_mlflow: mlruns
 experiment_name: ???
 
 # setup
-scaling_cfg: ../../configs/ShuffleMergeSpectral_trainingSamples-2_files_0_50.json # for DataLoader initialisation
+scaling_cfg: ../../configs/ShuffleMergeSpectral_trainingSamples-2_rerun_files_0_19.json # for DataLoader initialisation
```
kandrosov, line 13: I suggest using the latest file `ShuffleMergeSpectral_trainingSamples-2_files_0_498.json`. Or it has some known issues?

kandrosov (changes requested):

Training/python/2018v1/train.yaml
shedprog: @lguzzi Is ShuffleMergeSpectral_trainingSamples-2_files_0_498.json scaling for the ShuffleMergeSpectral_trainingSamples-2_rerun version of S&M?

lguzzi: yes, they were run on the rerun version of the ntuples

PR #87, opened 2022-02-25
Boosted Tau Production Changes for Ultra Legacy

aloeliger:
After some particularly circuitous commits adding some hacks and then removing them, this is the current status of the boosted tau effort UL production. The only real changes after the added and removed hacks is the introduction of some sample types and global tags, and a change to the `isRun2UL` function. For Boosted Tau convenience I have added a quick filter that can be added into the production that simply checks that there is some boosted tau in the collection before attempting to do any Ntuplization, but if this is not desirable in the main repository, it can be removed from the commit.

We will be attempting to make ntuples from all 4 UL time periods under this setup.

I will open a PR for the general boosted tau changes, but that similarly needs to be cleaned up to prevent superfluous merge conflicts from distracting from genuine ones.

### DIFF
--- a/Production/plugins/BoostedTauProductionFilter.cc
+++ b/Production/plugins/BoostedTauProductionFilter.cc
@@ -1,0 +1,156 @@
+// system include files
+#include <memory>
+
+// user include files
+#include "FWCore/Framework/interface/Frameworkfwd.h"
+#include "FWCore/Framework/interface/stream/EDFilter.h"
+
+#include "FWCore/Framework/interface/Event.h"
+#include "FWCore/Framework/interface/MakerMacros.h"
+
+#include "FWCore/ParameterSet/interface/ParameterSet.h"
+#include "FWCore/Utilities/interface/StreamID.h"
+#include "DataFormats/PatCandidates/interface/Tau.h"
+
+//
+// class declaration
+//
+
+class BoostedTauProductionFilter : public edm::stream::EDFilter<> {
+   public:
+      explicit BoostedTauProductionFilter(const edm::ParameterSet&);
+      ~BoostedTauProductionFilter();
+
+      static void fillDescriptions(edm::ConfigurationDescriptions& descriptions);
+
+   private:
+      virtual void beginStream(edm::StreamID) override;
+      virtual bool filter(edm::Event&, const edm::EventSetup&) override;
+      virtual void endStream() override;
+
+      //virtual void beginRun(edm::Run const&, edm::EventSetup const&) override;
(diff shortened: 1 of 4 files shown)


Production/plugins/BoostedTauProductionFilter.cc
```diff
@@ -0,0 +1,156 @@
+// system include files
```
kandrosov, line 1: is this filter still needed?

Production/python/sampleConfig.py
```diff
@@ -41,6 +45,10 @@
                  'Emb_18D' : '102X_dataRun2_Prompt_v15',
                  'MC_Phase2_110X' : '110X_mcRun4_realistic_v3',
                  'MC_Phase2_111X' : 'auto:phase2_realistic_T15',
+                 'UL18': '106X_upgrade2018_realistic_v15_L1v1',
+                 'UL17': '106X_mc2017_realistic_v8',
+                 'UL16': '106X_mcRun2_asymptotic_v17',
+                 'UL16APV': '106X_mcRun2_asymptotic_preVFP_v11'
```
kandrosov, line 51: - there are already `MC_UL18` and `RunUL18` in the list. Please, remove duplicates and follow the naming notation
- data and MC have different GTs
- use GTs recommended in this twiki https://twiki.cern.ch/twiki/bin/view/CMS/PdmVRun2LegacyAnalysis

kandrosov (changes requested):

kandrosov (changes requested):

Production/python/Production.py
```diff
@@ -115,7 +115,15 @@
...
-    boostedTaus_InputTag = cms.InputTag('slimmedTausBoosted')
+    from TauMLTools.Production.runTauIdMVA import runTauID
+    updatedBoostedTauName = "slimmedBoostedTausNewID"
+    runTauID(process, outputTauCollection=updatedBoostedTauName, inputTauCollection="slimmedTausBoosted",
+             toKeep = [ "2017v2", "dR0p32017v2", "newDM2017v2", "deepTau2017v2p1" ])
+    process.boostedSequence = cms.Sequence(
+        getattr(process, updatedBoostedTauName+'rerunMvaIsolationSequence') *
+        getattr(process, updatedBoostedTauName)
+    )
+    boostedTaus_InputTag = cms.InputTag(updatedBoostedTauName)
```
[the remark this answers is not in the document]
aloeliger: I tried running Ntuples without these lines, the MVA branches ended up empty (all 0), and it was causing issues with validation (I need to rerun a few ntuples now). These lines remedy the problem.

PR #88, opened 2022-03-02
TensorFlow DataSet Compatibility + other bug fixes

lucasrussell01:
- Introduced the possibility of reading data from previously saved tf datasets for training
- Added a script ROOT_to_tf to save ROOT files as a tf dataset
Minor fixes:
- Conversion from torch to tf in get_predict_generator fro apply training
- Change size-t -> long int in C++ parser to avoid problem observed when DeepTauVSjet cut set to -1
- Fixed small problem for file names with dashes for evaluation
- Made histo2d get weights shared pointer in definition
- Fix bug in regularisation implementation for convolutional layers
- Improved error messages when NaN found in tensor

### DIFF
--- a/Training/interface/DataLoader_main.h
+++ b/Training/interface/DataLoader_main.h
@@ -190,10 +190,10 @@ public:
         input_histogram .th2d_add(*(input_th2d .get()));
 
         target_histogram.divide(input_histogram);
-        hist_weights[tau_type] = std::make_shared<TH2D>(target_histogram.get_weights_th2d(
+        hist_weights[tau_type] = target_histogram.get_weights_th2d(
             ("w_1_"+tau_name).c_str(),
             ("w_1_"+tau_name).c_str()
-        ));
+        );
         if (debug) hist_weights[tau_type]->SaveAs(("Temp_"+tau_name+".root").c_str()); // It's required that all bins are filled in these histograms; save them to check incase binning is too fine and some bins are empty
 
         target_histogram.reset();
--- a/Training/interface/DataLoader_main.h
+++ b/Training/interface/DataLoader_main.h
@@ -776,7 +776,7 @@ public:
             fillGrid(Br::muon_dphi, DeltaPhi(tau.muon_phi.at(idx), tau.tau_phi));
 
             fillGrid(Br::muon_dxy, tau.muon_dxy.at(idx));
-            if(tau.muon_dxy_error.at(idx) != 0)
+            if(std::isnormal(tau.muon_dxy_error.at(idx)) && std::isnormal(tau.muon_dxy.at(idx))) 
               fillGrid(Br::muon_dxy_sig, std::abs(tau.muon_dxy.at(idx)) / tau.muon_dxy_error.at(idx));
 
             const bool normalizedChi2_valid = tau.muon_normalizedChi2.at(idx) >= 0;
--- a/Training/python/2018v1/ROOT_to_tf.py
+++ b/Training/python/2018v1/ROOT_to_tf.py
@@ -1,0 +1,47 @@
+import os
+import yaml
+import gc
+import sys
+from glob import glob
+import math
+import numpy as np
+import tensorflow as tf
+os.environ["CUDA_VISIBLE_DEVICES"]="-1"
+sys.path.insert(0, "..")
+from common import *
+import DataLoader
+import argparse
+
+parser = argparse.ArgumentParser(description='Convert ROOT files to TF dataset')
+parser.add_argument('--n_batches', required=True, type=int, help="Number of batches")
+parser.add_argument('--scaling_cfg', required=True, type=str, help="Scaling config")
+parser.add_argument('--save_path', required=True, type=str, help="Save path")
+parser.add_argument('--training_cfg', required=True, type=str, help="Training config")
+args = parser.parse_args()
+
+
+
+save_path = args.save_path # "/home/russell/tfdata/testing"
+scaling_cfg = args.scaling_cfg #"../../configs/ShuffleMergeSpectral_trainingSamples-2_files_0_50.json"
+training_cfg_path = args.training_cfg #../../configs/training_v1.yaml
+
+with open(training_cfg_path) as file:
+    training_cfg = yaml.full_load(file)
+    print("Training Config Loaded")
+
+training_cfg["SetupNN"]["n_batches"]=args.n_batches
+training_cfg["SetupNN"]["n_batches_val"]=0 # only generate training data as train/val split done later in training
+training_cfg["SetupNN"]["validation_split"]=0
+training_cfg["Setup"]["input_type"]="ROOT" # make ROOT so generator loads
+
+
+dataloader = DataLoader.DataLoader(training_cfg, scaling_cfg)
+print("DataLoader Created")
+gen_train = dataloader.get_generator(primary_set = True, return_weights = dataloader.use_weights, show_progress=True) 
+print("Generator Loaded")
+input_shape, input_types = dataloader.get_input_config()
+print("Input shapes and Types acquired")
+data_train = tf.data.Dataset.from_generator(gen_train, output_types = input_types, output_shapes = input_shape).prefetch(tf.data.AUTOTUNE)
+print("Dataset extracted from DataLoader")
+tf.data.experimental.save(data_train, save_path, compression = "GZIP")
+print("Conversion Complete")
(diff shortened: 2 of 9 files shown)


Training/python/2018v1/Training_v0p1.py
```diff
@@ -267,16 +290,38 @@ def compile_model(model, opt_name, learning_rate):
...
+            my_ds = ds
+        cell_locations = data_loader.cell_locations
+        active_features = data_loader.active_features
+        active = [] #list of elements to be kept
+        if "TauFlat" in active_features:
+            active.append(1)
+        if "inner" in cell_locations:
+            active.extend([2,3,4])
+        if "outer" in cell_locations:
+            active.extend([5,6,7])
```
shedprog, line 308: The order of the cell may not be static and depends on how we define it in config: https://github.com/cms-tau-pog/TauMLTools/blob/ed7b7b385219fa6a239bf972502cb8edf92754d1/Training/configs/training_v1.yaml#L72-L76
Better to read dynamically order of CellType from config or precompiled enume classes. Same corresponds to rm_inner function.

Training/python/DataLoader.py
```diff
@@ -118,6 +121,8 @@ def get_generator(self, primary_set = True, return_truth = True, return_weights
         def _generator():
 
             finish_counter = 0
+            if show_progress:
+                pbar = tqdm(total = n_batches)
```
shedprog, line 125: `n_batches` is not always a reasonable number, e.g: It is set to -1 when we are reading all dataset.

Training/python/DataLoader.py
```diff
@@ -164,6 +171,7 @@ def get_predict_generator(self):
         '''
         assert self.batch_size == 1
         data_loader = R.DataLoader()
+        converter = torch_to_tf(return_truth=True, return_weights=False)
```
shedprog, line 174: I propose to add this args to get_predict_generator signature and read it from there

Training/interface/DataLoader_main.h
```diff
@@ -190,10 +190,10 @@ class DataLoader {
         input_histogram .th2d_add(*(input_th2d .get()));
 
         target_histogram.divide(input_histogram);
-        hist_weights[tau_type] = std::make_shared<TH2D>(target_histogram.get_weights_th2d(
+        hist_weights[tau_type] = target_histogram.get_weights_th2d(
```
shedprog, line 193: Is this change needed because of the scope?

shedprog (changes requested):

Training/python/2018v1/ROOT_to_tf.py
```diff
@@ -0,0 +1,47 @@
...
+
+dataloader = DataLoader.DataLoader(training_cfg, scaling_cfg)
+print("DataLoader Created")
+gen_train = dataloader.get_generator(primary_set = True, return_weights = dataloader.use_weights, show_progress=True) 
+print("Generator Loaded")
+input_shape, input_types = dataloader.get_input_config()
+print("Input shapes and Types acquired")
+data_train = tf.data.Dataset.from_generator(gen_train, output_types = input_types, output_shapes = input_shape).prefetch(tf.data.AUTOTUNE)
+print("Dataset extracted from DataLoader")
+tf.data.experimental.save(data_train, save_path, compression = "GZIP")
```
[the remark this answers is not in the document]
lucasrussell01: There's no option to limit the memory of this function, but it saves the files continuously, it doesn't wait until everything is loaded.

Training/python/DataLoader.py
lucasrussell01: Will implement this thanks, it's just a progress bar for when loading the dataset using tensorflow, so will disable it if n_batches<0

Training/python/DataLoader.py
lucasrussell01: Will also add this

Training/interface/DataLoader_main.h
lucasrussell01: Yes

Training/python/2018v1/Training_v0p1.py
lucasrussell01: @shedprog I've now added an argument tf_dataset_x_order, which will allow us to specify the order of the layers in the tf dataset we load, in case eg we disabled some when it was generated, and rm_inner and reshape_tensor now read the relevant indeces from this

Training/python/2018v1/Training_v0p1.py
```diff
@@ -267,16 +295,42 @@ def compile_model(model, opt_name, learning_rate):
...
+            my_ds = ds
+        cell_locations = data_loader.cell_locations
+        active_features = data_loader.active_features
+        active = [] #list of elements to be kept
+        if "TauFlat" in active_features:
+            active.append(tauflat_index)
+        if "inner" in cell_locations:
+            active.extend(inner_indices)
+        if "outer" in cell_locations:
+            active.extend(outer_indices)
```
shedprog, line 317: To my understanding the model input shape is pre-defined in the `netConf_full` variable, you can not just change the number of tensors in the input (to NN) tuple because it already considers fixed number and defined here: https://github.com/cms-tau-pog/TauMLTools/blob/ed7b7b385219fa6a239bf972502cb8edf92754d1/Training/python/DataLoader.py#L185-L219

Also my question related to the indexes of the tensor (em, muon, hadrons,...) was referring that the sequential number of this index is already defined in meta enume class (see attachments) [SetupMeta.txt](https://github.com/cms-tau-pog/TauMLTools/files/8237831/SetupMeta.txt)
if you at some point need this index, maybe it would be better by designed not to define `tf_dataset_x_order` but take it from `enum class CellObjectType {` for instance `getattr(R.CellObjectType,"PfCand_electron")` would give you 1 or if you need the order of the grouped tensors after DataLoading it is following this order: https://github.com/cms-tau-pog/TauMLTools/blob/ed7b7b385219fa6a239bf972502cb8edf92754d1/Training/configs/training_v1.yaml#L72-L76 if you remove from there one of the list it should not return this tensor. But in principle this part is not interfering with RUN3 training, so it is up to you.

lucasrussell01: From the testing I've done, changing the number of inputs in the tensors works if you change the parameters that I introduced to switch off either inner or outer cones (cell_locations) or the tau dense layer (removing it from active_features), this information is passed to net config. We then need to remove the concerned layers from the saved tensor if using tf input format as they are usually created with all layers active (eg generated with all 7 tensors 3xinner, 3xouter and tau dense, so if we want to then disable for example inner for a training from this tf dataset we need to remove the 3 inner tensors from the dataset as the model will only expect 4 tensors). Let me know if this doesn't make sense.

lucasrussell01: Will also have a look at the indexes you're referring to in SetupMeta, thanks!

Training/interface/DataLoader_main.h
kandrosov: the problem is that [get_weigths_th2d](https://github.com/cms-tau-pog/TauMLTools/blob/ed7b7b385219fa6a239bf972502cb8edf92754d1/Training/interface/histogram2d.h#L176) (as it is in master) created shared pointer inside, but returned a reference:
```c++
TH2D& Histogram_2D::get_weights_th2d(const char* name, const char* title) {
  ...
  histo_ = std::shared_ptr<TH2D>(new TH2D(name, title, nx, xmin_, xmax_, ny, yaxis_.front(), yaxis_.back()));
  return *histo_;
}
```

So the original `histo_` was destructed before the copy constructor is called...

kandrosov (changes requested):

lucasrussell01:
@shedprog No I have not yet found a fix for that, this was just a typo I happened to find

Training/python/2018v1/Training_v0p1.py
```diff
@@ -303,7 +303,19 @@ def run_training(model, data_loader, to_profile, log_suffix):
...
+            if n_inner % 2 == 0 or n_outer % 2 == 0:
+                raise Exception("Number of cells not supported")
+            inner_size = data_loader.inner_cell_size
+            outer_size = data_loader.outer_cell_size
+            n_inner_right = (n_inner - 1) / 2
+            n_outer_right = np.ceil(n_inner_right * inner_size /  outer_size)
+            n_outer_to_remove = n_outer_right*2 + 1
+            i_middle = (n_outer-1)/2
+            i_start = int(i_middle - (n_outer_to_remove-1)/2)
+            i_end = int(i_middle + (n_outer_to_remove-1)/2 + 1) # +1 as end index not included
```
kandrosov, line 317: `(n_outer_to_remove-1)/2 == n_outer_right` => remove `n_outer_to_remove` and use `n_outer_right`

lucasrussell01: @kandrosov Good point, fixed now

PR #93, opened 2022-04-14
Few eval changes proposal

shedprog:
Not urgent PR, but would be nice to stabilize some aspects of evaluate performance scripts if possible, unless it is of course very local whims

Here I add / request few small changes to make life easier:
1) Not to run many times apply_training, we can loop over files in the folder.
2) If I don't apply training for all files in the folder e.g: for /eos/../DYJetsToLL_M-50-amcatnloFXFX_ext2/* I should be able to run just on the files I produced, for this in loops we need to rely only on available *.h5 files not on all *.root files.
3) Propose to disable checkout_train_repo by default because it can play a dangerous trick when stash the changes and then not pop it back if script fails.
4) There are some problems with plot_roc.py, when I put current training as a reference then I had an overlap of colors in the 1.0 line
5) To do: is it possible to define axis and ranges on the plot_roc.py level, because it hard to rerun evaluate_performance for an every cosmetic changes? (now all this dropped in performance.json on the evaluate_performance step)

### DIFF
--- a/Evaluation/apply_training.py
+++ b/Evaluation/apply_training.py
@@ -2,6 +2,7 @@ import os
 import sys
 import json
 import git
+import glob
 from tqdm import tqdm
 
 import uproot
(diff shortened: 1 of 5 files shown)


Evaluation/apply_training.py
```diff
@@ -58,67 +59,79 @@ def main(cfg: DictConfig) -> None:
...
+        output_filename = os.path.splitext(os.path.basename(input_file_name))[0]+"_pred"  if cfg.input_filename is None \
+                          else cfg.output_filename
+        if os.path.exists(f'{path_to_artifacts}/predictions/{cfg.sample_alias}/{output_filename}.h5'):
+            print("File exists: ", f'{path_to_artifacts}/predictions/{cfg.sample_alias}/{output_filename}.h5')
+            continue
+
+        # open input file
+        with uproot.open(input_file_name) as f:
+            t = f['taus']
+            n_taus = len(t['evt'].array())
```
kandrosov, line 78:
```suggestion
            n_taus = f['taus'].numentries
```

Evaluation/plot_roc.py
```diff
@@ -41,7 +41,11 @@ def __init__(self, data, ref_roc=None, WPcurve=True):
 
         if ref_roc is None:
```
ofivite, line 42: I just realised that the colour overlap at 1.0 line might be happening because of this if statement on lines 42-43 (since it essentially always adds 1.0 line on the ratio, while it is useless). Would it resolve the problem if these lines are removed?

While disabling ratio plotting for WP-like curves can be a quick fix in the very short term, I think it should be considered in general. We also had quite an extensive discussion with @DennRoy in #79 on how to include this use case and I'd say it's worth keeping it.

ofivite (commented):
Sorry for reviewing this late. In general I like and fully agree with your additions, leaving some comments and suggestions inline as a minor follow-up.

PR #94, opened 2022-04-15
Implement a possibility to resume training from the existing one

shedprog:
At the moment, just load weights option seems does not work, because DeepTauModel() was implemented after I started the training. compile_loss - also is not needed since loss now is hardcoded in the training loop, but it is needed for previous train version  

P.S: I don't understand how to link PR to the project point, maybe lucking permissions

### DIFF
--- a/Training/python/2018v1/Training_v0p1.py
+++ b/Training/python/2018v1/Training_v0p1.py
@@ -26,6 +26,7 @@ mlflow.tensorflow.autolog(log_models=False)
 import hydra
 from hydra.utils import to_absolute_path
 from omegaconf import DictConfig, OmegaConf
+import json
 
 sys.path.insert(0, "..")
 from common import *
@@ -365,7 +366,7 @@ def create_model(net_config, model_name):
     model = DeepTauModel(input_layers, softmax_output, name=model_name)
     return model
 
-def compile_model(model, opt_name, learning_rate):
+def compile_model(model, loss, opt_name, learning_rate):
     # opt = keras.optimizers.Adam(lr=learning_rate)
     opt = getattr(tf.keras.optimizers, opt_name)(learning_rate=learning_rate)
     #opt = tf.keras.optimizers.Nadam(learning_rate=learning_rate, schedule_decay=1e-4)
@@ -379,11 +380,15 @@ def compile_model(model, opt_name, learning_rate):
         TauLosses.Hcat_eInv, TauLosses.Hcat_muInv, TauLosses.Hcat_jetInv,
         TauLosses.Fe, TauLosses.Fmu, TauLosses.Fjet, TauLosses.Fcmb
     ]
-    model.compile(loss=None, optimizer=opt, metrics=metrics, weighted_metrics=metrics) # loss is now defined in DeepTauModel
+    model.compile(loss=loss, optimizer=opt, metrics=metrics, weighted_metrics=metrics) # loss is now defined in DeepTauModel
 
     # log metric names for passing them during model loading
     metric_names = {(m if isinstance(m, str) else m.__name__): '' for m in metrics}
     mlflow.log_dict(metric_names, 'input_cfg/metric_names.json')
+
+    # need to save only custome object for model load
+    metric_names_cus = {(m.__name__): '' for m in metrics if not isinstance(m, str)}
+    mlflow.log_dict(metric_names_cus, 'input_cfg/metric_names_custom.json')
 
 def run_training(model, data_loader, to_profile, log_suffix):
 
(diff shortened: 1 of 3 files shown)


ofivite:
I would actually like to discuss the mlflow logging in case of pretraining. Basically, now upon every execution of `Training_v0p1.py` script a new unique mlflow run is created within the specified `experiment_id` from yaml cfg:
https://github.com/cms-tau-pog/TauMLTools/blob/138d0523a63c973b1dcc10c1fa4e355ce18e84b1/Training/python/2018v1/Training_v0p1.py#L474-L484

Therefore, if we re-execute training script in pretraining mode, all the params/files logs will be saved in a new mlflow run, in a usual way as it's also done for the original training. This is OK ofc, but I think it's possible to do bookkeeping slightly more elegant. 

There is a concept of nested runs (see [here](https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.start_run) in the example field), which roughly speaking defines a "parent" run and allows to create "child" runs within it. In mlflow UI it looks like this: 
image

Under the hood it still creates a new run under the same experiment, but from the user experience perspective I think it makes the logic more structured. So if we adopt this option for pretraining, it would be clearly visible in UI which runs were done "from scratch" (parent runs) and which runs are pretrained (child runs) and from which run (corresponding parent run).

Do you think it would make sense to adjust the current setup to incorporates this option? I briefly tried to create some nested runs and it looks rather simple and straightforward to implement.

kandrosov:
yes, having nested runs would be very nice.

ofivite:
ok, then I think it is sufficient (but definitely would be better to test it first) to substitute the current mlflow setup in the beginning of `main()` with:
```python
 experiment = mlflow.get_experiment_by_name(cfg.experiment_name) 
 if experiment is not None: # fetch existing experiment id 
     run_kwargs = {'experiment_id': experiment.experiment_id}
     if cfg["pretrained"] is not None: # initialise with pretrained run, otherwise create a new run
         run_kwargs['run_id'] = cfg["pretrained"]["run_id"]
 else: # create new experiment 
     experiment_id = mlflow.create_experiment(cfg.experiment_name) 
     run_kwargs = {'experiment_id': experiment_id} 
  
 # run the training with mlflow tracking 
 with mlflow.start_run(**run_kwargs) as main_run: 
     if cfg["pretrained"] is not None: # create a child run
         mlflow.start_run(experiment_id=run_kwargs['experiment_id'], nested=True)
     active_run = mlflow.active_run()
     run_id = active_run.info.run_id

# <code as before>

# in the very end:
     mlflow.end_run() # since child run wasn't initialised with "with" statement, it should be ended explicitly
```

Training/python/2018v1/Training_v0p1.py
```diff
@@ -491,8 +498,30 @@ def main(cfg: DictConfig) -> None:
         print("loss consts:",TauLosses.Le_sf, TauLosses.Lmu_sf, TauLosses.Ltau_sf, TauLosses.Ljet_sf)
 
         netConf_full = dataloader.get_net_config()
+        compile_loss = None if setup["loss"] is None else getattr(TauLosses,setup["loss"])
```
kandrosov, line 501: - why do we need `compile_loss`? As far as I see it should always be `None`
- make sure that `loss` specified in `yaml` is used during the gradient descent.

kandrosov (changes requested):

Training/python/2018v1/Training_v0p1.py
shedprog: I forgot to wrap it inside DeepTauModel, let me fix it. The problem is that we want to try different losses but as it is hard-coded inside DeepTauModel that is not very elegant

kandrosov: I think the best way is to pass the loss name to the `DeepTauModel` constructor.

Training/python/2018v1/Training_v0p1.py
```diff
@@ -480,7 +482,12 @@ def main(cfg: DictConfig) -> None:
         run_kwargs = {'experiment_id': experiment_id}
 
     # run the training with mlflow tracking
-    with mlflow.start_run(**run_kwargs) as active_run:
+    with mlflow.start_run(**run_kwargs) as main_run:
```
ofivite, line 485: Unless I am mistaken, with the current mlflow setup still a new parent run will be created and then a child run nested inside of it. While the original idea was that nested runs (corresponding to fine-tuning of a pretrained model), will be placed inside of an already existing run (corresponding to  the pretrained model).

This is because `run_kwargs` doesn't contain information about pretrained run to be started first as a parent run. So the corresponding case should be added to `run_kwargs` definition, for example like:

```python
 experiment = mlflow.get_experiment_by_name(cfg.experiment_name) 
 if experiment is not None: # fetch existing experiment id 
     run_kwargs = {'experiment_id': experiment.experiment_id}
     if cfg["pretrained"] is not None: # initialise with pretrained run, otherwise skip and create a new run
         run_kwargs['run_id'] = cfg["pretrained"]["run_id"]
 else: # create new experiment 
     experiment_id = mlflow.create_experiment(cfg.experiment_name) 
     run_kwargs = {'experiment_id': experiment_id} 

with mlflow.start_run(**run_kwargs) as main_run:
### ...
```

ofivite (changes requested):

kandrosov (approved):

PR #103, opened 2022-04-23
Netweork splitting 2p5

lguzzi:
This PR adds a script to split and deploy the 2p5 version of deeptau as frozen graphs.
The splitting is done as follows:

- inner model: starts from grid input and stops at last 1x1 convolution
- outer model: as inner
- core model: starts from flat features+inner output + outer output and stops at probabilities
conv. 1x1 layers have been replaced by dense layers acting on a single grid cell through the input features. This has required the reshaping of the inner/outer models inputs (11x11xF -> F - F=n. of input features) and outputs (11x11xM -> M - M = number of output features). @kandrosov Should we add and explicit reshape before the inner and outer outputs?

The scripts checks for nans and infs in the input model weights.
The script runs a comparison between the predictions of the full model (as loaded from the checkpoint file) and the created chunks. The predictions must differ by less than 1e-4 (floating precision - test runs OK on the current version).
The script then saves the split models into pb files (frozen graphs) using cmsml.

more details in the attached pdf.
[deeptau_splitting.pdf](https://github.com/cms-tau-pog/TauMLTools/files/8547527/deeptau_splitting.pdf)

can run using
```bash
python split_network_2p5.py --output test_deploy --scaling ../configs/ShuffleMergeSpectral_trainingSamples-2_files_0_498.json --training ../configs/training_v1.yaml  --input /eos/cms/store/group/phys_tau/TauML/prod_2018_v2/mlruns/4/7041fb6941b64392b51bcacc648e4da7/artifacts/model_checkpoints/full_v2_e4_step1_e2.tf
```
.

### DIFF
--- a/Training/python/split_network_2p5.py
+++ b/Training/python/split_network_2p5.py
@@ -105,10 +105,9 @@ def CoreModel(inner, outer):
     current_grid_size = net_config.n_cells[loc]
     n_inputs = inner.shape.as_list()[1] if loc == 'inner' else outer.shape.as_list()[1] #was 3 on conv 2d
     n = 1
-    lname = inner.name if loc=='inner' else outer.name
+    lname = inner.name.split('/')[0] if loc=='inner' else outer.name.split('/')[0]
     lshape = inner.shape.as_list()[1] if loc == 'inner' else outer.shape.as_list()[1]
-    prev_layer = Input(name=lname+'_split_input', shape=(net_config.n_cells[loc], net_config.n_cells[loc], lshape))
-    aliases[lname+'_split_input'] = ['input_'+loc]
+    prev_layer = Input(name=lname, shape=(net_config.n_cells[loc], net_config.n_cells[loc], lshape))
     input_layers.append(prev_layer)
     while current_grid_size > 1:
       win_size = min(current_grid_size, conv_2d_net_setup.window_size)
@@ -127,7 +126,6 @@ def CoreModel(inner, outer):
   final_dense = reduce_n_features_1d(features_concat, dense_net_setup, 'final')
   output_layer = Dense(net_config.n_outputs, name="final_dense_last", kernel_initializer=dense_net_setup.kernel_init)(final_dense)
   softmax_output = Activation("softmax", name="main_output")(output_layer)
-  aliases['main_output'] = 'main_output/Softmax'
   return keras.Model(input_layers, softmax_output, name='core')
 
 def ConvDenseAlias(loc):
@@ -165,13 +163,6 @@ def ConvDenseAlias(loc):
 
   return keras.Model(input_layers, prev_layer, name=loc)
 
-def apply_aliases(model):
-  global aliases
-  for k, v in aliases.items():
-    if not k in [l.name for l in model.layers]:
-      continue
-    model.get_layer(k)._name = v
-
 full_model = LoadModel(args.input)
 
 inner_model = ConvDenseAlias('inner')
@@ -194,10 +185,6 @@ with open(args.output+"/core_summary.txt", "w") as smr:
   core_model.summary(print_fn=lambda x: smr.write(x+'\n'))
 with open(args.output+"/full_summary.txt", "w") as smr:
   full_model.summary(print_fn=lambda x: smr.write(x+'\n'))
-
-apply_aliases(inner_model)
-apply_aliases(outer_model)
-apply_aliases(core_model)
 
 save_to_graph(inner_model, args.output)
 save_to_graph(outer_model, args.output)
--- a/Training/python/split_network_2p5.py
+++ b/Training/python/split_network_2p5.py
@@ -47,8 +47,8 @@ def test_prediction(model, reference):
     m_shape = tuple(model.inputs[i].shape.as_list()[1:])
     if p_shape==m_shape:
       i_pred.append(pred)
-    elif len(p_shape)==3 and len(m_shape)==1 and p_shape[2]==m_shape[0]:
-      i_pred.append(pred.reshape((p_shape[0]*p_shape[1], p_shape[2])))
+    elif len(p_shape)==len(m_shape)==3 and p_shape[2]==m_shape[2]:
+      i_pred.append(pred.reshape((p_shape[0]*p_shape[1], 1, 1, p_shape[2])))
     else:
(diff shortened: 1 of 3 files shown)


kandrosov (changes requested):
- To be compatible with DeepTau v2p1, input shapes for inner/outer models should be as here https://github.com/cms-sw/cmssw/blob/master/RecoTauTag/RecoTau/plugins/DeepTauId.cc#L1761-L1772
- call `save_to_graph` also for the full model

Training/python/split_network_2p5.py
```diff
@@ -142,11 +142,12 @@ def ConvDenseAlias(loc):
   for comp_id in range(len(net_config.comp_names)):
     comp_name = net_config.comp_names[comp_id]
     n_comp_features = net_config.n_comp_branches[comp_id]
-    input_layer_comp = Input(name="input_{}_{}".format(loc, comp_name), shape=n_comp_features)
+    input_layer_comp = Input(name="input_{}_{}".format(loc, comp_name), shape=(1,1,n_comp_features))
     input_layers.append(input_layer_comp)
+    input_layer_resh = keras.layers.Reshape(name='{}_reshape'.format(input_layer_comp.name), target_shape=(n_comp_features,))(input_layer_comp)
```
kandrosov, line 147: Reshape should not be needed. A dense layer should act on the last dimension by default. But please, check it, in case something is changed in the last TF versions.

kandrosov (changes requested):

Training/python/split_network_2p5.py
```diff
@@ -0,0 +1,196 @@
...
+  outer_model.summary(print_fn=lambda x: smr.write(x+'\n'))
+with open(args.output+"/core_summary.txt", "w") as smr:
+  core_model.summary(print_fn=lambda x: smr.write(x+'\n'))
+with open(args.output+"/full_summary.txt", "w") as smr:
+  full_model.summary(print_fn=lambda x: smr.write(x+'\n'))
+
+for layer in core_model.inputs:
+  layer._name = 'input_inner' if 'inner' in layer.name else 'input_outer' if 'outer' in layer.name else layer.name 
+assert 'input_inner' in [l.name for l in core_model.inputs] and 'input_outer' in [l.name for l in core_model.inputs], "Core model input layer names are wrong"
+
```
ofivite, line 190: If this slash postfix matters, then it would be good also to rename the output layers (dropout) for inner/outer models to `{inner/outer}_all_dropout_4/Identity` (compatibility with [DeepTauId.cc#L1730](https://github.com/cms-tau-pog/cmssw/blob/5b56ea4f85b845ed3238c59310f79185ff481ab5/RecoTauTag/RecoTau/plugins/DeepTauId.cc#L1730))

ofivite (changes requested):
Just minor renamings from my side

Training/python/split_network_2p5.py
ofivite: Indeed I tested this briefly and it seems to be fine without reshaping: 

image

 However, reshaping with the current setup won't be compatible with [DeepTauId.cc#L1833](https://github.com/cms-tau-pog/cmssw/blob/5b56ea4f85b845ed3238c59310f79185ff481ab5/RecoTauTag/RecoTau/plugins/DeepTauId.cc#L1833) (if I understood `features.tensor<float, 4>()` operation correctly), so it should be either another reshaping done in the end of the inner/outer model, or the initial one removed.

lguzzi: Hi, yes, reshape is not needed. I added it because it makes weight copying easiier

Training/python/split_network_2p5.py
```diff
@@ -0,0 +1,196 @@
...
+    high_level_features.append(processed_tau)
+
+  # here we need the original names still, so that the test_prediciton will not fail
+  for loc in ['inner', 'outer']:
+    current_grid_size = net_config.n_cells[loc]
+    n_inputs = inner.shape.as_list()[1] #was 3 on conv 2d
+    n = 1
+    lname = inner.name if loc=='inner' else outer.name
+    lshape = inner.shape.as_list()[1] if loc == 'inner' else outer.shape.as_list()[1]
+    prev_layer = Input(name=lname+'_split_input', shape=(net_config.n_cells[loc], net_config.n_cells[loc], lshape))
```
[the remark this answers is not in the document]
lguzzi: This is done and checked in https://github.com/lguzzi/TauMLTools/blob/split2p5/Training/python/split_network_2p5.py#L187-L189
In the lines above, I create input layers with name "original_name/something_split_input". Doing so, I can easily get the layer "original_name" from the full model with get_layer("original_name"), which makes testing the split model in ```test_prediction```easier.

Training/python/split_network_2p5.py
```diff
@@ -0,0 +1,196 @@
...
+      n_inputs = n_filters
+    
+    cells_flatten = Flatten(name="{}_cells_flatten".format(loc))(prev_layer)
+    high_level_features.append(cells_flatten)
+  
+  features_concat = Concatenate(name="features_concat", axis=1)(high_level_features)
+  dense_net_setup.ComputeLayerSizes(features_concat.shape.as_list()[1])
+  final_dense = reduce_n_features_1d(features_concat, dense_net_setup, 'final')
+  output_layer = Dense(net_config.n_outputs, name="final_dense_last", kernel_initializer=dense_net_setup.kernel_init)(final_dense)
+  softmax_output = Activation("softmax", name="main_output")(output_layer)
```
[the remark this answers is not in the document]
lguzzi: See the comments in the previous section.
https://github.com/cms-tau-pog/TauMLTools/pull/103/commits/867d8ff27cadb8841165fac4952d4e11f84df666 applies also here

Training/python/split_network_2p5.py
lguzzi: see previous two sections

Training/python/split_network_2p5.py
lguzzi: I removed the reshape layer in https://github.com/cms-tau-pog/TauMLTools/pull/103/commits/2a5c54f53ad6a9aab11cbcd7d214b6487101bbb5

kandrosov (approved):

PR #109, opened 2022-05-04
Updated Phase 2 configurations, and my implementation of a new loss function

DennRoy:


Training/python/Phase2/Training.py
```diff
@@ -33,11 +33,16 @@
...
+    def __init__(self, *args, loss=None, n_tau=None, **kwargs):
         super().__init__(*args, **kwargs)
         self.loss_tracker = keras.metrics.Mean(name="loss")
         self.pure_loss_tracker = keras.metrics.Mean(name="pure_loss")
         self.reg_loss_tracker = keras.metrics.Mean(name ="reg_loss")
+        if loss is None:
+            self.model_loss = TauLosses.tau_crossentropy_v2
+        else:
+            self.model_loss = getattr(TauLosses,loss)
+        self.n_tau = n_tau
```
kandrosov, line 45: no need to pass n_tau to the constructor use `n_tau = tf.shape(x[0])[0]` inside train/test step

Training/python/Phase2/Training_WPloss.py
```diff
@@ -36,11 +36,16 @@
 
 class DeepTauModel(keras.Model):
```
kandrosov, line 37: would it be possible to have a single `Training_CNN.py` shared between phase 2 and run 3 and set parameters (including the choice of the loss function) from the config?

kandrosov (changes requested):

Training/python/Phase2/Training.py
DennRoy: This doesn't seem to work, at least not without running TF with eagerly. `tf.shape(x[0])[0]` returns `Tensor("strided_slice:0", shape=(), dtype=int32)`.
It gives me the error: `TypeError: x and y must have the same dtype, got tf.float32 != tf.int32`. (Though strangely, it works whether I make n_tau an actual int for float.)
And when trying to convert to int (or float): `TypeError: int() argument must be a string, a bytes-like object or a number, not 'Tensor'`

kandrosov: Are you using `tf.cast` when you trying to convert int to float? E.g.
```python
n_tau = tf.cast(tf.shape(x[0])[0], x[0].dtype)
```

DennRoy: No I wasn't, but it does work like this, thanks!

Training/python/Phase2/Training_WPloss.py
DennRoy: Ok, I removed the second file, and now the regular `Training.py` can be used for both the tau_crossentropy and the new loss.
One note however: I was never able to make the new loss function work without using eagerly evaluation, so now it's enabled only when using the new loss function:
https://github.com/cms-tau-pog/TauMLTools/blob/fb754c3f85a087563ac7c351080291bcd5474be9/Training/python/Phase2/Training.py#L619

kandrosov: Can we merge code even further? What are the differences between the python code for Phase 2 and Run 3? As far as I see, there is no reason to diverge at the code level - I would keep the same code for Run 3, boosted and Phase 2, with the different configuration files for each case.

DennRoy: You mean to merge `2018v1/Training_v0p1.py` and `Phase2/Training.py`? There are a few differences, which I can try to merge. What I notice at first glance is that the metrics for tensorboard are different (the list is hardcoded), because when using the new loss function with eagerly evaluation, reducing the number of metric helps to somewhat compensate for the longer runtime.

kandrosov: yes, we can rename the merged file to `Training_CNN.py`. A list of active metrics can be moved to the config - it shouldn't be a problem.

Training/python/Phase2/Training.py
```diff
@@ -31,6 +31,231 @@
...
+            for typ,ID in zip(["e", "mu", "jet"], [0,1,3]):
+                self.res[typ] = tf.where(self.model.y[:,ID]==1.0, self.model.y_pred[:,2]/(self.model.y_pred[:,2]+self.model.y_pred[:,ID]), -1)
+                self.res["tau"+typ] = tf.where(self.model.y[:,2]==1.0, self.model.y_pred[:,2]/(self.model.y_pred[:,2]+self.model.y_pred[:,ID]), -1)
+            self.res["w"] = self.model.sample_weight
+            self.firstbatch = False
+        else:
+            for typ,ID in zip(["e", "mu", "jet"], [0,1,3]):
+                self.res[typ] = tf.concat([self.res[typ], tf.where(self.model.y[:,ID]==1.0, self.model.y_pred[:,2]/(self.model.y_pred[:,2]+self.model.y_pred[:,ID]), -1)], axis=0)
+                self.res["tau"+typ] = tf.concat([self.res["tau"+typ], tf.where(self.model.y[:,2]==1.0, self.model.y_pred[:,2]/(self.model.y_pred[:,2]+self.model.y_pred[:,ID]), -1)], axis=0)
+            self.res["w"] = tf.concat([self.res["w"], self.model.sample_weight], axis=0)
```
shedprog, line 163: Can you point which part of the code requires eager execution, it is this one?

Training/interface/DataLoader_main.h
```diff
@@ -535,6 +536,8 @@ class DataLoader {
               fillGrid(Br::pfCand_ele_dxy_sig, std::abs(tau.pfCand_dxy.at(pfCand_idx)) / tau.pfCand_dxy_error.at(pfCand_idx));
               fillGrid(Br::pfCand_ele_dz, tau.pfCand_dz.at(pfCand_idx));
               fillGrid(Br::pfCand_ele_dz_sig, std::abs(tau.pfCand_dz.at(pfCand_idx)) / tau.pfCand_dz_error.at(pfCand_idx));
+              fillGrid(Br::pfCand_ele_time, tau.pfCand_time.at(pfCand_idx));
+              fillGrid(Br::pfCand_ele_time_sig, std::abs(tau.pfCand_time.at(pfCand_idx)) / tau.pfCand_timeError.at(pfCand_idx));
```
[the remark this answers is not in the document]
DennRoy: I was able to run a test training for Run3, after adding all variables to the `Features_disabled` list:
https://github.com/cms-tau-pog/TauMLTools/blob/97a90a10597c04149c101ebe18ba23c1c3463e77/Training/configs/training_v1.yaml#L483-L487
But I was also wondering if this `isfinite` check found here:
https://github.com/cms-tau-pog/TauMLTools/blob/97a90a10597c04149c101ebe18ba23c1c3463e77/Training/interface/DataLoader_main.h#L635
should be applied to all inputs, inside the `fillGrid` call. Would that work?

Training/python/Phase2/Training.py
DennRoy: Yes, that's right. I save the `y` and `y_pred` from `train_step` to the class and read them here. `self.model.y`, `self.model.y_pred` and `self.model.sample_weight` are some unreadable format without eager execution.

Training/interface/DataLoader_main.h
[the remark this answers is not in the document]
shedprog: This disabling of the features means that their enume index is = -1 and this variable will not pass this condition here: https://github.com/cms-tau-pog/TauMLTools/blob/97a90a10597c04149c101ebe18ba23c1c3463e77/Training/interface/DataLoader_main.h#L467-L471
and will not be added to the tensor, however, the division itself std::abs(tau.pfCand_time.at(pfCand_idx)) / tau.pfCand_timeError.at(pfCand_idx) will be done before getting inside the function. I am afraid that tau.pfCand_timeError.at(pfCand_idx) can just point to some garbage in memory, and this garbage can be just 32 bits of zeros.

I propose in addition to `if(std::isfinite(tau.pfCand_dz.at(pfCand_idx))){ ` add non-zero denominator check.

Training/python/Phase2/Training.py
shedprog: > Yes, that's right. I save the `y` and `y_pred` from `train_step` to the class and read them here. `self.model.y`, `self.model.y_pred` and `self.model.sample_weight` are some unreadable format without eager execution.

I see, but this format should be tensorflow tensor, and all operations here are tf.ops (since you are not using THists anymore), there should be tf-friendly way to accumulate some output values

kandrosov: I would advise against using `model.sample_weight`, `model.y` etc. inside callbacks.
Another point is that you concatenate all outputs in memory, which is not scalable for big datasets. Aggregation should be used already inside `on_train_batch_end`.
I think implementing per-batch code as a metric would fit more general TF design concepts.
E.g. metric will aggregate nFakes and nTaus for each WP and then callback will process them in `on_epoch_end` (they will be available through `log` argument).

DennRoy: Thanks for the idea! But I can't just save the number of Taus/Fakes passing the current WPs, I need all discriminator values to update the WPs for the next iteration.
I can do this step in a metric instead (so that doesn't require eager), but is there a way to directly access the weights inside the metric too?

kandrosov: yes, `metric.update_state` takes weights as an argument. You can even call it directly if you want to have better control over how it is called. E.g. as it is done for some adversarial metrics
https://github.com/cms-tau-pog/TauMLTools/blob/ef7c8a9971a4da590907fe4af839588e32f56dee/Training/python/2018v1/Training_v0p1.py#L133

DennRoy: Actually it seems that moving this implementation to a metric doesn't work either. The code for the metric is only called once before the epoch, not after every batch (At least, that's what it seems like based on some printouts). So the tensors I try to fill end up being `Tensor("SelectV2:0", shape=(None,), dtype=float32)` after the epoch.

kandrosov: it is correct - there should be only one (ok, for some cases two) printout using standard python `print` when the graph is built, and then the graph is executed for each batch (no more python code calls). To confirm it you can put printout in the `train_step`. To have a printout at every real call, [tf.print](https://www.tensorflow.org/api_docs/python/tf/print) should be used.

You'll need to define your custom metric (see example [here](https://www.tensorflow.org/api_docs/python/tf/keras/metrics/Metric)) and declare all variables to keep n_fake, n_true counts etc with `self.add_weight`. Then in `update_state` update these variables with `assign_add` or similar methods.

DennRoy: I found some (old?) code on such an implementation:
https://github.com/cms-tau-pog/TauMLTools/blob/master/Training/python/commonReco.py#L58-L81
And I tried setting up the metric like this. But I seem unable to define 1-dim tensor variables to store the discriminators. There is always some error about shape mismatch, either like `ValueError: Shape must be rank 1 but is rank 0 for '{{node Reshape}} = Reshape[T=DT_FLOAT, Tshape=DT_INT32](SelectV2, Reshape/shape)' with input shapes: [?], []` or like `ValueError: Cannot assign value to variable ' TauDiscr0:0': Shape mismatch.The variable shape (1,), and the assigned value shape () are incompatible.`
I'm also not sure if this implementation would work either, because it looks like `assign_add` only adds new values on top of exists one, instead of appending them?

kandrosov:
@DennRoy any progress in unifying Run3 and Phase 2 training code?

DennRoy:
From my side it should be all done, except for the issue where I can't figure out a solution to have the new loss function with updated WPs work without eager execution. The unified Training_CNN.py is here: https://github.com/cms-tau-pog/TauMLTools/pull/109/commits/97a90a10597c04149c101ebe18ba23c1c3463e77
If you're okay with merging this, I'd just have to figure out how to resolve this one merge conflict.

DennRoy:
Should the old files really be removed? Personally I'd prefer to make an "Archive" directory and put everything in there (including the `2017v2` and `reco` directories)

kandrosov:
> Should the old files really be removed? Personally I'd prefer to make an "Archive" directory and put everything in there (including the `2017v2` and `reco` directories)

We have tag for [DeepTau v2](https://github.com/cms-tau-pog/TauMLTools/tree/deepTau_v2) training. I've also created a tag for [DeepTau v2p5](https://github.com/cms-tau-pog/TauMLTools/tree/deepTau_v2p5) training. I think the better strategy is to keep only the current version of the code in the master branch and use git (tags, commits history etc.) to preserve and navigate through the old developments. But we can discuss it today at the meeting to collect all opinions.

Training/python/Training_CNN.py
```diff
@@ -96,6 +96,7 @@ def run_pred():
                 reg_loss = reg_losses # Empty
                 loss = pure_loss
                 loss_vec = pure_loss_vec
+            self.y_pred = y_pred
```
kandrosov, line 99: do not store `y_pred` in self

Training/python/Training_CNN.py
```diff
@@ -184,6 +188,10 @@ def test_step(self, data):
             self.compiled_metrics.update_state(y, y_pred_class, sample_weight)
         # Return a dict mapping metric names to current value
         metrics_out = {m.name: m.result() for m in self.metrics}
+        self.x = x
+        self.y = y
+        self.sample_weight = sample_weight
+        self.y_pred = y_pred
```
kandrosov, line 194: do not store `x`, `y`, `sample_weight`, `y_pred` in self

Training/python/reco/Training_SNNv0.py
```diff
@@ -1,394 +0,0 @@
-import os
```
kandrosov, line 1: do not delete reco files

Training/python/reco/Training_test.py
```diff
@@ -1,32 +0,0 @@
-import os
```
kandrosov, line 1: do not delete reco files

Training/python/reco/train.yaml
```diff
@@ -1,16 +0,0 @@
-defaults:
```
kandrosov, line 1: do not delete reco files

kandrosov (changes requested):

DennRoy:
Done; I realized I didn't need to store `x`, or the other ones either in `test_step`, but `y`, `y_pred` and `sample_weight` in `train_step` are required for the new loss function. So now it's only stored when actually using the new loss function.

kandrosov (approved):

PR #112, opened 2022-05-10
Update crab_submit.py

lguzzi:
fix a small bug preventing the correct arbitration between python 2 and 3


shedprog:
This fix works, I guess we can merge this PR

PR #116, opened 2022-05-23
WP derivation

ofivite:
This PR adds a [`derive_wp.py`](https://github.com/cms-tau-pog/TauMLTools/compare/master...ofivite:dev?expand=1#diff-261ac3eeea02a1ebbc0944fea04970dc5cecaa839b2403705e6fdf271425a9db) script to compute working points given the model predictions. The code is an adapted and cleaned up version of the previously used [`EfficiencyEstimation.ipynb`](https://github.com/cms-tau-pog/TauMLTools/blob/master/Training/notebooks/2017v2/EfficiencyEstimation.ipynb). All the necessary configuration is specified in the corresponding [`derive_wp.yaml`](https://github.com/cms-tau-pog/TauMLTools/compare/master...ofivite:dev?expand=1#diff-90deff0ab9ff98901cfccbc7198a263ae7db2a1de1d3e16825fc51a14828cbaa) config file and the computation follows the logic:
* `WPMaker` class steers the computation
* Take genuine taus from `create_df.pred_samples` within `create_df.pred_path` (by default assumed to be within mlflow artifacts, but can be specified differently) which pass `create_df.selection`. Feature values for selection will be taken from corresponding to every pred file input files, retrieved from `pred_input_filemap.json` (assumed to be present in `pred_path/{sample_name}`)
* `score_vs_{type}` is computed for each tau as usual (`score_tau/(score_tau+score_vs_type)`)
* Function `update_thrs()`  for the each `score_vs_{type}` computes thresholds corresponding to its quantiles given by `tpr` values. `tpr` is defined as a grid of evenly-spaced values with `step=wp_maker.tpr_step`
* Thresholds which yield efficiencies closest to the values in `wp_maker.wp_definitions.{type}.wp_eff` are selected
* Procedure is repeated until convergence of threshold values for all `vs_types`. If `wp_maker.require_wp_vs_others=False`, it should take only 2 iterations. Otherwise, the dependancy on passing the loosest WPs vs remaining tau types is introduced in the definition of WP vs `current_type`, so convergence will require >2 iterations (for example below it should be 7 iterations).  

Example of running on lxplus:
```bash
python derive_wp.py create_df._path_to_mlflow=/eos/cms/store/group/phys_tau/TauML/prod_2018_v2/mlruns create_df._experiment_id=4 create_df._run_id=5371f5d7080846c1a213f0e648471c11
```

Lastly and not related to WP computation, small rearrangement is made to `yaml` configs in `Evalution` folder, so that they are all placed inside `configs` folder.


Evaluation/configs/derive_wp.yaml
```diff
@@ -3,31 +3,53 @@ wp_maker:
...
   pred_samples: # samples in `path_to_preds` to take taus from 
     GluGluHToTauTau_M125: 'eventTuple_1-7_pred.h5' # single files, list of files and "*" are supported 
   
   # paths to input ROOT files will be retrieved from corresponding pred_input_filemap.json
   input_branches: [ 'tau_pt', 'tau_eta', 'tau_decayMode' ]
   input_tree_name: taus
-  selection: '(tau_pt>=20) and (tau_pt<100)'
-  tau_types: ['e', 'mu', 'tau', 'jet']
\ No newline at end of file
+  selection: '(tau_pt>=20) and (tau_pt<100) and not (tau_decayMode in [5,6])'
```
kandrosov, line 55: sorry, I forgot, aslo `abs(tau_dz) < 0.2` should be applied

Evaluation/configs/derive_wp.yaml
```diff
@@ -0,0 +1,33 @@
...
+  _run_id: ???
+  path_to_preds: "${create_df._path_to_mlflow}/${create_df._experiment_id}/${create_df._run_id}/artifacts/predictions/"
+  pred_samples: # samples in `path_to_preds` to take taus from 
+    GluGluHToTauTau_M125: 'eventTuple_1-7_pred.h5' # single files, list of files and "*" are supported 
+  
+  # paths to input ROOT files will be retrieved from corresponding pred_input_filemap.json
+  input_branches: [ 'tau_pt', 'tau_eta', 'tau_decayMode' ]
+  input_tree_name: taus
+  selection: '(tau_pt>=20) and (tau_pt<100)'
+  tau_types: ['e', 'mu', 'tau', 'jet']
```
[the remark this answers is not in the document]
ofivite: It was needed internally to create "tau vs type" scores for each type in `tau_types` except for "tau". But you are right, its presence in the config is redundant and confusing , so I removed it in 446c6c4e4bbbcc04961f94c905ce4d3457d12ab5

Evaluation/configs/derive_wp.yaml
```diff
@@ -0,0 +1,33 @@
+wp_maker:
+  _target_: derive_wp.WPMaker
+
+  wp_definitions:
+    e:
+      wp_names: ["VVTight", "VTight", "Tight", "Medium", "Loose", "VLoose", "VVLoose", "VVVLoose"]
+      wp_eff: [0.6, 0.7, 0.8, 0.9, 0.95, 0.98, 0.99, 0.995]
```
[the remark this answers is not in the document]
ofivite: done in 4c0b152e2501355cc2ff0cb47de99a4c6231a125, I also added an option to specify which WPs should be passed (previously it was fixed to be the loosest ones)

kandrosov (changes requested):

Evaluation/configs/derive_wp.yaml
ofivite: actually, what about `abs(tau_eta) < 2.3`, I think it also should be applied at this point?

kandrosov: Yes, indeed :)

kandrosov (approved):

PR #122, opened 2022-07-08
Update to Phase 2 configs, and "fix" to never load NANs in DataLoader

DennRoy:
- Update of Phase 2 configurations, and also the configurations to be used when taking v2p5 and continuing the training on Phase 2 samples.
- And perhaps more importantly: A fix to never load NANs in DataLoader. This is needed because our Phase 2 ntuples have a few inputs (Photon track information) which are NAN for very few taus (less than 5 I believe). For the regular Phase 2 trainings, we had these inputs disabled, but they are necessary when continuing the training from v2p5, because the inputs are enabled there. So, instead, I added a "fix" in the DataLoader to make sure that any input is finite before is it actually read. I do want to make sure if this implementation is fine for you? Because this means that it won't catch any future instances of NANs (perhaps only in previous steps of the Shuffle&Merge or features scaling workflows).

### DIFF
--- a/Training/configs/training_phase2.yaml
+++ b/Training/configs/training_phase2.yaml
@@ -68,7 +68,7 @@ SetupNN:
     n_batches_val        : -1 # -1 for all
     n_batches_log        : 1000 # monitor every n batches in tensorboard
     epoch                : 0
-    n_epochs             : 15
+    n_epochs             : 10
     validation_split     : 0.2
     max_queue_size       : 1
     n_load_workers       : 1
@@ -77,7 +77,7 @@ SetupNN:
                             [ GridGlobal, PfCand_muon, Muon ], # muons
                             [ GridGlobal, PfCand_chHad, PfCand_nHad ] # hadrons
                            ]
-    TauLossesSFs         : [0.4, 1.0, 2.0, 0.6]
+    TauLossesSFs         : [0.8, 1.5, 4.0, 1.7]
     optimizer_name       : "Nadam"
     learning_rate        : 0.001
     schedule_decay       : 0.0 #0.0001 # Using step decay instead of schedule decay
--- a/Training/interface/DataLoader_main.h
+++ b/Training/interface/DataLoader_main.h
@@ -472,7 +472,7 @@ public:
                                     std::make_index_sequence<nFeaturesTypes>{});
 
         auto fillGrid = [&](auto _feature_idx, float value) {
-          if(static_cast<int>(_feature_idx) < 0) return;
+          if((static_cast<int>(_feature_idx) < 0) || !(std::isfinite(value))) return;
           const CellObjectType obj_type = FeaturesHelper<decltype(_feature_idx)>::object_type;
           const size_t start = start_indices.at(ElementIndex<decltype(_feature_idx), FeatureTuple>::value);
           data->x_grid.at(obj_type).at(inner).at(start + static_cast<int>(_feature_idx))
--- a/Training/python/train_phase2_contv2p5.yaml
+++ b/Training/python/train_phase2_contv2p5.yaml
@@ -1,0 +1,25 @@
+hydra:
+  run:
+    dir: outputs/${experiment_name}/${training_cfg.SetupNN.model_name}/${now:%Y-%m-%d_%H-%M-%S}
+defaults:
+  - ../configs@_global_.training_cfg: training_phase2_contv2p5
+  - _self_
+
+# mlflow
+path_to_mlflow: mlruns
+experiment_name: ???
+
+# pretrained: null
+pretrained:
+  run_id : e1f3ddb3c4c94128a34a7635c56322eb # e.g: 6076d0ebeb414e38b29e5511e7f7d3ab
+  experiment_id: 4 # e.g: 1
+  starting_model: adv_decay_1e3_Adam_100k_step1_final.tf # e.g: DeepTau2018v0_step1_final.tf
+
+# setup
+scaling_cfg: ../configs/ScalingParameters_Phase2_contv2p5.json # for DataLoader initialisation
+gpu_cfg:
+  gpu_mem  : 7 # in Gb
+  gpu_index: 0
+
+# logs
+log_suffix: step1
--- a/Training/configs/ScalingParameters_Phase2_contv2p5.json
+++ b/Training/configs/ScalingParameters_Phase2_contv2p5.json
@@ -1,0 +1,3252 @@
+{
+    "TauFlat": {
+        "rho": {
+            "global": {
+                "mean": 100.0,
+                "std": 100.0,
+                "lim_min": -1.0,
+                "lim_max": 1.0,
+                "sqmean": null,
+                "num": null
+            }
+        },
+        "tau_pt": {
+            "global": {
+                "mean": 510.0,
+                "std": 490.0,
+                "lim_min": -1.0,
+                "lim_max": 1.0,
+                "sqmean": null,
+                "num": null
+            }
+        },
+        "tau_eta": {
+            "global": {
+                "mean": 0.0,
+                "std": 3.0,
+                "lim_min": -1.0,
+                "lim_max": 1.0,
+                "sqmean": null,
+                "num": null
+            }
+        },
+        "tau_phi": {
+            "global": {
+                "mean": 0.0,
+                "std": 3.141592653589793,
+                "lim_min": -1.0,
+                "lim_max": 1.0,
+                "sqmean": null,
+                "num": null
+            }
+        },
+        "tau_mass": {
+            "global": {
+                "mean": 0.5718,
+                "std": 0.5081,
+                "lim_min": -5,
+                "lim_max": 5,
+                "sqmean": 0.5851,
+                "num": 710346
+            }
+        },
+        "tau_E_over_pt": {
+            "global": {
+                "mean": 2.181,
+                "std": 1.593,
+                "lim_min": -5,
+                "lim_max": 5,
+                "sqmean": 7.292,
+                "num": 710346
+            }
+        },
+        "tau_charge": {
+            "global": {
+                "mean": 0,
+                "std": 1,
+                "lim_min": "-inf",
+                "lim_max": "inf",
+                "sqmean": null,
+                "num": null
+            }
+        },
+        "tau_n_charged_prongs": {
+            "global": {
+                "mean": 0,
+                "std": 1,
+                "lim_min": "-inf",
+                "lim_max": "inf",
+                "sqmean": null,
+                "num": null
+            }
+        },
+        "tau_n_neutral_prongs": {
+            "global": {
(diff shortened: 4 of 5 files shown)


Training/configs/training_phase2.yaml
```diff
@@ -3,13 +3,12 @@ Setup:
     iso_cone             :  0.5
     n_inner_cells        :  11
     inner_cell_size      :  0.02
-    n_outer_cells        :  21
-    outer_cell_size      :  0.05
+    n_outer_cells        :  25
+    outer_cell_size      :  0.042
```
kandrosov, line 7: Do we want to have `n_outer_cells=25` as the baseline for Phase 2? If yes, could you add to the PR description a link to the slides that justify the increase of the number of cells?

Training/interface/DataLoader_main.h
```diff
@@ -472,7 +472,7 @@ class DataLoader {
                                     std::make_index_sequence<nFeaturesTypes>{});
 
         auto fillGrid = [&](auto _feature_idx, float value) {
-          if(static_cast<int>(_feature_idx) < 0) return;
+          if((static_cast<int>(_feature_idx) < 0) || !(std::isfinite(value))) return;
```
kandrosov, line 475: I think throwing an exception is a better way to handle non-finite values. At the level of training it is desirable to have control of the inputs, i.e. we should know the domain of each input (including when it is not valid) and treat this cases before calling `fillGrid`.

Training/python/train_phase2_contv2p5.yaml
```diff
@@ -0,0 +1,25 @@
...
+experiment_name: ???
+
+# pretrained: null
+pretrained:
+  run_id : e1f3ddb3c4c94128a34a7635c56322eb # e.g: 6076d0ebeb414e38b29e5511e7f7d3ab
+  experiment_id: 4 # e.g: 1
+  starting_model: adv_decay_1e3_Adam_100k_step1_final.tf # e.g: DeepTau2018v0_step1_final.tf
+
+# setup
+scaling_cfg: ../configs/ScalingParameters_Phase2_contv2p5.json # for DataLoader initialisation
```
kandrosov, line 19: please add comment here describing why the original v2p5 scaling yaml can't be used (e.g. the list of variables for which scaling has changed)

kandrosov (changes requested):

Training/interface/DataLoader_main.h
kandrosov: Do we have many variables that can have unexpected NaN values? I hope that the amount of variables is rather limited and we can treat each one separately to not loose "physics sense" why variable is undefined. E.g. if there is no track details, it is expected that `dxy_error` is not defined etc. If we start blindly screening all the variables, we risking to mis some errors in the input dataset or misuse some inputs.

Indeed, there is an exception thrown in the later parts of the code, I wasn't sure if it is still enabled by default. So then we shouldn't have check here.

DennRoy: Looking at it again (and testing successfully with the other Scaling config), it's only for `pfCand_gamma_vertex_dz`/`_tauFL`. But even "blinding" these two inputs, wouldn't that cause the same issues in the example you're giving? Would it just be better to leave this as a comment in the code, so that one may locally make changes to skip NANs in specific scenarios such as this, where it's known that there are just a few taus with NANs?

kandrosov: The question is from where do you know it? If NaN screeening would be in place, would you have noticed that the variables in question sometimes ill-defined?
IMO, we should have a good knowledge of the variable domain which we are sure is up-to-date. For the future R&D, we should consider reduction of the "useless" or ill-defined input variables. E.g. it is very likely that the vertex information for pf gammas can be easily dropped without any performance losses.

Training/interface/DataLoader_main.h
```diff
@@ -472,7 +472,7 @@ class DataLoader {
                                     std::make_index_sequence<nFeaturesTypes>{});
 
         auto fillGrid = [&](auto _feature_idx, float value) {
-          if(static_cast<int>(_feature_idx) < 0) return;
+          if(static_cast<int>(_feature_idx) < 0) return; // To skip NANs, in cases where a few are known to exist: if((static_cast<int>(_feature_idx) < 0) || !(std::isfinite(value))) return;
```
kandrosov, line 475: If there are some `pfCand_gamma` variables with NaN which are not screened, it should be done in the dataLoader code. We can have a single if block for all track-related pf gamma variables. E.g.
```c++
if(tau.pfCand_vertex_z.at(pfCand_idx) != tau.pv_z) {
  fillGrid(Br::pfCand_gamma_vertex_dx, tau.pfCand_vertex_x.at(pfCand_idx) - tau.pv_x);
  fillGrid(Br::pfCand_gamma_vertex_dy, tau.pfCand_vertex_y.at(pfCand_idx) - tau.pv_y);
  fillGrid(Br::pfCand_gamma_vertex_dz, tau.pfCand_vertex_z.at(pfCand_idx) - tau.pv_z);
  fillGrid(Br::pfCand_gamma_vertex_dt, tau.pfCand_vertex_t.at(pfCand_idx) - tau.pv_t);
  fillGrid(Br::pfCand_gamma_vertex_dx_tauFL, tau.pfCand_vertex_x.at(pfCand_idx) - tau.pv_x -
                                             tau.tau_flightLength_x);
  fillGrid(Br::pfCand_gamma_vertex_dy_tauFL, tau.pfCand_vertex_y.at(pfCand_idx) - tau.pv_y -
                                             tau.tau_flightLength_y);
  fillGrid(Br::pfCand_gamma_vertex_dz_tauFL, tau.pfCand_vertex_z.at(pfCand_idx) - tau.pv_z -
                                             tau.tau_flightLength_z);
} else {
  fillGrid(Br::pfCand_gamma_vertex_dx_tauFL, tau.pv_x - tau.tau_flightLength_x);
  fillGrid(Br::pfCand_gamma_vertex_dy_tauFL, tau.pv_y - tau.tau_flightLength_y);
  fillGrid(Br::pfCand_gamma_vertex_dz_tauFL, tau.pv_z - tau.tau_flightLength_z);
}
```

kandrosov (changes requested):

Training/interface/DataLoader_main.h
DennRoy: I believe it's in the feature scaling step, where I first encountered those NANs. The data are loaded differently in that step, right? So I thought that, yes, I would know beforehand.

Training/python/train_phase2_contv2p5.yaml
```diff
@@ -0,0 +1,25 @@
...
+experiment_name: ???
+
+# pretrained: null
+pretrained:
+  run_id : e1f3ddb3c4c94128a34a7635c56322eb # e.g: 6076d0ebeb414e38b29e5511e7f7d3ab
+  experiment_id: 4 # e.g: 1
+  starting_model: adv_decay_1e3_Adam_100k_step1_final.tf # e.g: DeepTau2018v0_step1_final.tf
+
+# setup
+scaling_cfg: ../configs/ScalingParameters_Phase2.json # for DataLoader initialisation
```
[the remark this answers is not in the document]
DennRoy: Yes, I thought it makes sense to use the scaling parameters for Phase 2, right? Because those same scalings are used in the application step, and some parameters in Phase 2, like `rho` in particular, have a very different range.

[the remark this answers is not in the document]
kandrosov: ok, I see. Thanks for the confirmation!

Training/interface/DataLoader_main.h
DennRoy: I'm wondering if it's really necessary to ignore the `dx`, `dy`, `dt` varables too if only `dz` is bad (probably if you can't trust those values in that case?), but I guess it doesn't matter because it's only for a few taus anyway.
One comment: `a != b` will always be true if `a` and/or `b` are NAN, so I'll check with `isfinite` instead.

kandrosov: For quite NaNs, result of all boolean operators is `false`. E.g. `if(!(x == x))` is a standard way to check for NaNs. But ok, `isfinite` is more readable indeed :)

kandrosov (approved):

Training/interface/DataLoader_main.h
ofivite: If I may add (quite a long) comment, not fully relevant to the ongoing discussion but rather with a purpose of broadening the perspective a bit and possibly stepping aside from our current workflow. So if you'd like we can discuss this later at some meeting.

I generally think that it is a good practise to understand the input data, otherwise one is prone to have "garbage in, garbage out" results. It is a broad question what it means "to understand the data", but if we firstly consider a simpler NaN problem, it looks to me that the approach "if any number is NaN -> mask it" brings less understanding of the input domain then "if training crashed because of some NaN  -> find NaNed feature and mask its values in DataLoader manually". It might be annoying, because reaching the convergence requires to go throw several crashed trainings, but it forces the user to have a look at the distributions of input features and therefore gain the understanding. That is why I tend to agree with Konstantin about avoiding "inclusive" masking and trying to be specific.

However, I've just realised that fundamentally, this approach is also careless in terms of understanding the data, because it only makes to think about the data when there is a problem. A more consistent approach would be for the user to spend some time investigating all the distributions prior to running the training. But this approach suffers from the curse of dimensionality since we have quite a lot of features to look through and that takes time. Also, it is hard to visualise a dataset if it doesn't fit into the memory.

Ideally, it would be nice to have some kind of dashboard which shows at once all the histograms for the specified dataset/input feature with all necessary statistics about it (and I bet that there is already existing solutions which do that). So the current compromise to automate this "understanding" procedure within our framework is the feature scaling step, at least that's what I was thinking of when I was working on it.
 
I had an idea that since we iterate over the _whole_ dataset anyway, why don't we collect some knowledge as we go, aside from the scaling parameters? One example would be [scanning for NaN/inf](https://github.com/cms-tau-pog/TauMLTools/blob/31d2a54717657453d80436ec8ef16504d428faa4/Training/python/feature_scaling.py#L144-L149), but then there is [`plot_quantile_ranges.py`](https://github.com/cms-tau-pog/TauMLTools/blob/master/Training/python/plot_quantile_ranges.py) which job is to plot the quantile ranges for each feature -> help to understand its domain. And also [`plot_scaling_convergence.py`](https://github.com/cms-tau-pog/TauMLTools/blob/master/Training/python/plot_scaling_convergence.py) which should show if there are outliers in the convergence procedure -> point to anomalous features. 

So in principle, after running the feature scaling step and fully studying its outputs one can proceed to the masking of the features  in DataLoader (either exclusively or inclusively) which it showed to be NaNed, whilst at the same time having more confidence in understanding the input data. There is the caveat though that currently feature scaling preprocessing isn't automatically in synch with DataLoader preprocessing, which would be of course good to have unified across the framework.

PR #124, opened 2022-07-16
Adversarial ML Tools + documentation

lucasrussell01:
This PR contains scripts used for the adversarial training dataset production and evaluation that was used in v2p5, as well as some documentation. 
Added files
- "EventDrop.py" can be used to drop events from ROOT files to avoid large weights for very poorly balanced spectra.
- "EventMix.py" and "DataMixer.h" are used to mix the data and MC contributions together to generate the ROOT adversarial datasets.
- "AdversarialGenerate_tf.py" calls DataLoader with the new "GenerateAdversarial" option to assign pT weights and data/MC labels to taus when converting adversarial dataset to tensorflow format.
- "apply_adversarial_training.py" evaluates models on the adversarial testing dataset and saves the Djet scores for data/MC events (and y_adv for adversarial trained models) in the artifacts
- Jupyter notebook "evaluate_adversarial" to plot the adversarial predictions
- README_adversarial with some brief documentation of the steps


Training/configs/training_v1.yaml
```diff
@@ -20,14 +20,16 @@ Setup:
...
-    input_type           : "ROOT" # supported "Adversarial", 'ROOT' or 'tf'
+    input_type           : "ROOT" # supported: 'ROOT', 'tf', 'Adversarial', 'AdversarialGenerate'
     tf_input_dir         : /eos/cms/store/group/phys_tau/TauML/prod_2018_v2/tf_datasets/emb_21k_99cut # path to saved tf dataset
     tf_dataset_x_order   : ["TauFlat", "inner_egamma", "inner_muon", "inner_hadron", "outer_egamma", "outer_muon", "outer_hadron"] # specify order of layers in tf dataset
     adversarial_dataset  : "/eos/cms/store/group/phys_tau/TauML/prod_2018_v2/adversarial_datasets/AdvTraining" # False if not use, ds name if use
     adv_parameter        : [1, 10] # k1, k2 for gradients in common layers
     n_adv_tau            : 100 # number of candidates per adversarial batch
     adv_learning_rate    : 0.01
     use_previous_opt     : False
+    adversarial_weights  : "/home/russell/AdversarialTauML/TauMLTools/adv_weights.root" # weights for adversarial dataset generation
```
kandrosov, line 31: move weights to the central eos area and fix the path accordingly

Training/interface/DataLoader_main.h
```diff
@@ -192,7 +205,7 @@ class DataLoader {
       if (!target_th2d) throw std::runtime_error("Target histogram could not be loaded");
 
       for( auto const& [tau_type, tau_name] : tau_types_names)
-      {
+      { if (tau_type != 8){ // No histogram for data samples
```
kandrosov, line 208: - fix indentations 
- use enum instead of 8 (also below)

Training/interface/DataLoader_main.h
```diff
@@ -320,6 +343,13 @@ class DataLoader {
 
       }
 
+      const double GetAdversarialWeight(const ULong64_t dataset_id, const double pt) const
```
kandrosov, line 346:
```suggestion
      const double GetAdversarialWeight(const ULong64_t dataset_id, const double pt) const
```
```suggestion
      double GetAdversarialWeight(ULong64_t dataset_id, double pt) const
```

Training/interface/EventDrop.h
```diff
@@ -0,0 +1,81 @@
+#include "TFile.h"
```
kandrosov, line 1: a lot of code duplication between `EventDrop.h` and `DataMixer.h`. I suggest adding option to the `DataMixer` whatever to mix or to drop events and unify the code base

Training/python/EventMix.py
```diff
@@ -0,0 +1,103 @@
+import ROOT
```
kandrosov, line 1: `EventMix.py` -> `AdversarialEventMix.py`

Training/python/EventMix.py
```diff
@@ -0,0 +1,103 @@
...
+n_tt_jets = 1
+n_DY_muons = 1
+
+print("Mix Characteristics:")
+
+
+dataDesc = ROOT.DatasetDesc()
+dataDesc.n_per_batch = n_data
+dataDesc.selection = "data"
+dataDesc.fileNames.push_back("/home/russell/skimmed_tuples/MuTau_prod2018/SingleMuon_skimmed.root")
```
kandrosov, line 34: move input root files to the cenral eos storage and fix paths accordingly (here and below)

kandrosov (changes requested):

Training/interface/EventDrop.h
lucasrussell01: I have added an option so that only `DataMixer.h` is necessary, please let me know if you don't like the way I've implemented it or have any suggestions for improvement

Training/interface/DataLoader_main.h
lucasrussell01: I've now avoided using 8 and match to "data", by using tau_name for the first case and the unordered map find for the second, assuming this is ok

Training/interface/DataLoader_main.h
```diff
@@ -204,23 +204,23 @@ class DataLoader {
...
-            ("w_1_"+tau_name).c_str(),
-            ("w_1_"+tau_name).c_str()
-        );
-        if (debug) hist_weights[tau_type]->SaveAs(("Temp_"+tau_name+".root").c_str()); // It's required that all bins are filled in these histograms; save them to check incase binning is too fine and some bins are empty
-
-        target_histogram.reset();
-        input_histogram .reset();
-      } 
+      for( auto const& [tau_type, tau_name] : tau_types_names){
+          if (tau_name != "data"){ // No histogram for data samples
```
kandrosov, line 208: here and in similar places below: don't use strings (string operation are much slower than integer comparison) -> use enum values

Training/interface/DataMixer.h
```diff
@@ -1,15 +1,14 @@
...
     size_t n_per_batch;
-    std::string selection;
+    std::string selection; // selection for mixing
+    TH1D threshold; // threshold for event dropping
     std::vector<std::string> fileNames;
 };
 
-struct Dataset {
-    Dataset(const DatasetDesc& desc)  : desc_(desc) {
+struct DatasetMix {
```
kandrosov, line 10: use polymorphism
```c++
struct Dataset {
  virtual ~Dataset() {}
  virtual bool FillNext(input_tuple::TauTuple& outputTuple) = 0;
};

struct DatasetMix : Dataset {
  ...
  virtual bool FillNext(input_tuple::TauTuple& outputTuple) override
  {
    ...
  }
};

struct DatasetDrop : Dataset {
  ...
  virtual bool FillNext(input_tuple::TauTuple& outputTuple) override
  {
    ...
  }
};


class DataMixer {
public:
  DataMixer(const std::vector<DatasetDesc>& datasets_desc, const std::vector<OutputDesc>& outputDescs,
            const std::string& action)
    : outputDescs_(outputDescs)
  {
    if (action == "Mix"){
      CreateDatasets<DatasetMix>(datasets_desc);
    } else if(action == "Drop") {
      CreateDatasets<DatasetDrop>(datasets_desc);
    } else {
      throw std::runtime_error("Unsupported action");
    }
  }

  ...

private:
  template<typename DS>
  void CreateDatasets(const std::vector<DatasetDesc>& datasets_desc)
  {
    for(const auto& desc : datasets_desc) {
      std::shared_ptr<Dataset> ds = std::make_shared<DS>(desc);
      datasets_.push_back(ds);
    }
  }

private:
  std::vector<std::shared_ptr<Dataset>> datasets_;
};
```

kandrosov (changes requested):

kandrosov (approved):

Issue #115, opened 2022-05-16
Improvements to eval & plotting steps


ofivite:
As was mentioned in PR https://github.com/cms-tau-pog/TauMLTools/pull/93, at the moment:
- [ ] interface of reading input paths at prediction step is halfway to be fully within mlflow ecosystem ->  this needs to be completed

Also, things to be improved in the plotting workflow: 
- [x] enabling WP plotting
- [x] fixing ratio=1 overlapping problem
- [x] (if time permits) add flexibility to adjust axis range at plotting step
- [ ] updating README

PR #127, opened 2022-08-29
Update to evaluation workflow

ofivite:
This PR addresses the items from https://github.com/cms-tau-pog/TauMLTools/issues/115 related to the plotting routine. The updates are:
* Plotting setup is now moved from `evaluate_performance.py` step and can be flexibly configured at `plot_roc.py` step, globally as well as on per curve basis. 
* Duplication of each `RocCurve` and `PlotSetup` classes is removed and their methods are clarified. 
* Improved configurability of `create_df()` and rounding in WP computation. 
* Option to require WP in selection of `evaluate_performance.py` from the corresponding column (in input files), in addition to the current thresholds option.
* Enabling plotting of up/down uncertainties to WPs in ROC curves.
* Minor refactoring.
* Updated README, including removed deprecated descriptions.


shedprog (approved):
+1

maybe we also add customization, not to plot ration on the roc_curves?

ofivite:
> +1
> 
> maybe we also add customization, not to plot ration on the roc_curves?

In the sense that to have an option of showing only the upper panel (with ROC curves) or on the curve-by-curve basis, whether a given curve should be added to the ratio panel?

shedprog:
> 



> > +1
> > maybe we also add customization, not to plot ration on the roc_curves?
> 
> In the sense that to have an option of showing only the upper panel (with ROC curves) or on the curve-by-curve basis, whether a given curve should be added to the ratio panel?

Yes, like you do here on the second slide: https://indico.cern.ch/event/1183099/contributions/4970422/attachments/2482355/4261681/TaT_e1.pdf

PR #128, opened 2022-09-16
Adding photons to tau jet cones

msessini:


Production/plugins/TauTupleProducer.cc
```diff
@@ -96,6 +97,37 @@ namespace tau_analysis {
...
+        static float hcalDepth2OverEcal(const Gamma* photon) { return photon->hadronicDepth2OverEm(); }
+        static float hcalDepth1OverEcalBc(const Gamma* photon) { return photon->hadTowDepth1OverEm(); }
+        static float hcalDepth2OverEcalBc(const Gamma* photon) { return photon->hadTowDepth2OverEm(); }
+        static float full5x5_hcalDepth1OverEcal(const Gamma* photon) { return -999.; }
+        static float full5x5_hcalDepth2OverEcal(const Gamma* photon) { return -999.; }
+        static float full5x5_hcalDepth1OverEcalBc(const Gamma* photon) { return -999.; }
+        static float full5x5_hcalDepth2OverEcalBc(const Gamma* photon) { return -999.; }
+    };
+
+
```
shedprog, line 130: @msessini please resolve the existing conflict in this part of the code.

And add overloading for version == 11 like we do for Electrons:
```c++
    template<typename Elec>
    struct GetElecVer<Elec, 11> : GetElecVer<Elec, 10> {};
```

shedprog (changes requested):

Production/plugins/TauTupleProducer.cc
shedprog: It would make sense in general, but I still use UL for displaced taus. Also: what is the release for phase 2, isn't it 11_*_*?

Production/plugins/TauTupleProducer.cc
```diff
@@ -950,6 +1097,51 @@ class TauTupleProducer : public edm::EDAnalyzer {
...
+              tauTuple().sv_t.push_back(static_cast<float>(iSVtx.t()));
+	      tauTuple().sv_xE.push_back(static_cast<float>(std::sqrt(iSVtx.vertexCovariance(0,0))));
+              tauTuple().sv_yE.push_back(static_cast<float>(std::sqrt(iSVtx.vertexCovariance(1,1))));
+              tauTuple().sv_zE.push_back(static_cast<float>(std::sqrt(iSVtx.vertexCovariance(2,2))));
+              tauTuple().sv_tE.push_back(static_cast<float>(iSVtx.tError()));
+	      tauTuple().sv_chi2.push_back(static_cast<float>(iSVtx.vertexChi2()));
+              tauTuple().sv_ndof.push_back(static_cast<float>(iSVtx.vertexNdof()));
+	      for(int cand_idx : candsIdx) {
+		 tauTuple().sv_cands_svIdx.push_back(sv_idx);
+		 tauTuple().sv_cands_candIdx.push_back(cand_idx);
```
ofivite, line 1139: Currently only the indices of PF candidates which are assigned to `TauJet` object will be saved to `sv_cands_candIdx` (if they come from the associated SV). 

 Would it make sense to store all the PF indices which were used to reconstruct a given SV to have a more complete picture? For example, it could hint if we loose some candidates while applying the tau cone selection in `TauJet.cc`.

ofivite (changes requested):

Production/plugins/TauTupleProducer.cc
kandrosov: @shedprog - for UL it is safe to move to 12_4_X for miniAOD->TauTuple step. Actually we need to do this to include DeepTau v2p5 in the next ntuples.
@DennRoy - is it ok to move to >= 12_X for the future Phase 2 studies?

Production/plugins/TauTupleProducer.cc
kandrosov: yes, storing all candidates associated to SV makes sense. I think it would be better to include them at TauJet building stage to make sure that they are all stored in TauTuple.

Analysis/interface/TauTuple.h
```diff
@@ -50,6 +60,7 @@
     VAR(UInt_t, lumi) /* lumi section */ \
     VAR(ULong64_t, evt) /* event number */ \
     VAR(Int_t, npv) /* number of primary vertices */ \
+    VAR(Int_t, nsv) /* number of secondary vertices */ \
```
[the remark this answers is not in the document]
msessini: I forgot to fill it, it is done now.

Analysis/interface/TauTuple.h
```diff
@@ -336,16 +345,28 @@
...
-    PHOTON_VAR4(Float_t, e1x5, e2x5, e3x3, e5x5) /* shower shape signatures */ \
-    PHOTON_VAR3(Float_t, r1x5, r2x5, r9) /* shower shape signatures */ \
-    PHOTON_VAR(Float_t, maxEnergyXtal) \
-    PHOTON_VAR4(Float_t, hcalDepth1OverEcal, hcalDepth2OverEcal, hcalDepth1OverEcalBc, hcalDepth2OverEcalBc) /* HCAL over ECAL variables */ \
-    PHOTON_VAR2(Float_t, full5x5_sigmaEtaEta, full5x5_sigmaIetaIeta) /* shower shape spread */ \
-    PHOTON_VAR4(Float_t, full5x5_e1x5, full5x5_e2x5, full5x5_e3x3, full5x5_e5x5) /* shower shape signatures */ \
-    PHOTON_VAR3(Float_t, full5x5_r1x5, full5x5_r2x5, full5x5_r9) /* shower shape signatures */ \
-    PHOTON_VAR(Float_t, full5x5_maxEnergyXtal) \
-    PHOTON_VAR4(Float_t, full5x5_hcalDepth1OverEcal, full5x5_hcalDepth2OverEcal, full5x5_hcalDepth1OverEcalBc, full5x5_hcalDepth2OverEcalBc) /* HCAL over ECAL variables */ \
+    PHOTON_VAR2(Bool_t, passElectronVeto, hasPixelSeed) \
```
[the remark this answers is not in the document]
msessini: I am not fully aware either what these variables exactly mean, but yes I can do some research and comment a bit more.

Production/interface/Selectors.h
```diff
@@ -25,7 +25,7 @@ struct TauJetSelector {
 
     virtual ~TauJetSelector() {}
     virtual Result Select(const edm::Event& event, const std::deque<TauJet>& tauJets,
-                          const std::vector<pat::Electron>& electrons,
+                          const std::vector<pat::Electron>& electrons, const std::vector<pat::Photon>& photons,
```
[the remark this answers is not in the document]
msessini: Done.

Production/plugins/TauTupleProducer.cc
DennRoy: There are at least two changes required to make Phase2 production work with CMSSW 12. These are discussed here:
https://cms-talk.web.cern.ch/t/getting-muon-station-value-0/15765/1

msessini: @DennRoy, does CMSSW_12_3_6 seem right for phase2 production as replacement of CMSSW_11_2_5 ?

Analysis/interface/TauTuple.h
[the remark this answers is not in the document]
ofivite: Would be great, thanks! While searching for it myself I found this header file [DataFormats/EgammaCandidates/interface/Photon.h](https://github.com/cms-sw/cmssw/blob/6d2f66057131baacc2fcbdd203588c41c885b42c/DataFormats/EgammaCandidates/interface/Photon.h), which actually contains way more features. Do you know if they are also accessible through PAT::Photon and it makes sense to add them too?

Production/plugins/TauTupleProducer.cc
DennRoy: I'm not sure about any "best" release. I've used 12_4_0 and 12_4_8 before, which seems to have worked equally well. But I've also been told that 12_6_0_pre3 MIGHT fix the low tau reco efficiency in the Phase2 endcap (to be confirmed).

Analysis/interface/TauTuple.h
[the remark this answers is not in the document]
msessini: Hi @ofivite, these are in fact what the PHOTON_SHAPE variables in this code are. I added all the variables inside the ShowerShape structure, for both here [1]. On the other side, PHOTON_ variables are the ones from PAT::Photon.

https://github.com/cms-sw/cmssw/blob/6d2f66057131baacc2fcbdd203588c41c885b42c/DataFormats/EgammaCandidates/interface/Photon.h#L201-L202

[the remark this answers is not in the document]
ofivite: Yes indeed, but I was more referring to, for example, [`MIPVariables`](https://github.com/cms-sw/cmssw/blob/6d2f66057131baacc2fcbdd203588c41c885b42c/DataFormats/EgammaCandidates/interface/Photon.h#L324) or [`IsolationVariables `](https://github.com/cms-sw/cmssw/blob/6d2f66057131baacc2fcbdd203588c41c885b42c/DataFormats/EgammaCandidates/interface/Photon.h#L358) and there's other isolation variables there. Or did I misunderstand their meaning and it doesn't make sense (or possible) to include them?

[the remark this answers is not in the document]
kandrosov: - we have all PF candidates in the code, so isolation variables would be duplication (and create unnecessary dependency on photon iso algorithms).
- MIPVariables - I don't know how how they are computed. To be understood if it is safe for us to use them.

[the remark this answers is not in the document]
ofivite: OK I agree, let's go ahead with the current set of features and then reiterate if needed.
But it would be still nice to have a minimal features description, to avoid garbage in/garbage out path😊

Production/src/TauJet.cc
```diff
@@ -334,7 +360,8 @@ void TauJetBuilder::Build()
 
         for(size_t pfCandIndex = 0; pfCandIndex < cands_.size(); ++pfCandIndex) {
             const auto& pfCand = cands_.at(pfCandIndex);
-            if(!hasMatch(pfCand.polarP4())) continue;
+	    auto SVidx = collectSV(pfCand);
+            if(!hasMatch(pfCand.polarP4()) && SVidx<0) continue;
```
ofivite, line 364: Unless I'm mistaken, this condition implies that only PF candidates from SVs will be stored in `tauJet`, which isn't really what we want. So `SVidx < 0` should be removed.

Production/src/TauJet.cc
```diff
@@ -353,6 +380,8 @@ void TauJetBuilder::Build()
             if(tauJet.fatJet)
                 pfCandDesc.subJetDaughter = GetMatchedSubJetIndex(*tauJet.fatJet, pfCand);
             tauJet.cands.push_back(pfCandDesc);
+	    if(SVidx>0)
```
ofivite, line 383: Two things here: 
* I think vertex index can be 0 -> should it be `SVidx>=0`?
* I am not sure that the current data structure of storing SVs is intuitive and efficient. Correct me if I'm wrong, my understanding is that for each `tauJet` there's a collection of matched PF candidates required to come from SV (see comment above) and for _each_ of them associated SV is stored into a separate vector. I'd say that it will create many duplications, plus SVs currently do not carry information about PF candidates associated to it (see `FillSV()` function). That means it won't be possible to build the link between PF candidates and SVs afterwards.

So I would propose the following logic: for each `tauJet` object, store a collection of SVs, where every SVs is required to have at least one associated PF candidate coming from `tauJet`. Then, for all PF candidates in `tauJet` store an additional variable which is the index of associated SV (-1 if it doesn't exist). Alternatively, one can store for each SV a vector of associated PF candidates' indices. Does it sound reasonable?

ofivite (changes requested):

Production/src/TauJet.cc
msessini: This condition excludes PF candidates which are not in the cone and with no associated SV. However it keeps PF candidates inside and outside the cone whether they have a SV or not so previously stored candidates are still here.

ofivite: Ah right, sorry, I misunderstood the logic. Then it's fine👍🏻

Production/src/TauJet.cc
msessini: I agree, SVidx can be 0 thank you.

So the idea was to check for each PF candidate inside or outside the cone (as we discussed earlier) if it has a matched SV. If so, (i.e ```SVidx>=0```) we keep this SV in the collection. But you are right, if a SV has several associated PFCandidates it will be stored as many times. So maybe this piece of code can stay and I will include a small function in the ```FillSV()``` function of the TauTupleProducer that both check for duplicates and keep the associated PF candidate index in a vector, as you suggested.

ofivite: > Perhaps the fillSVMatched() function is useless and can be replaced by the line

Yes indeed, because `collectSV()` is already called before, so it is enough only to append the vertex with `SVidx` to `tauJet.secondVertices` (if `SVidx >=0`) 

> So maybe this piece of code can stay and I will include a small function in the FillSV() function of the TauTupleProducer that both check for duplicates and keep the associated PF candidate index in a vector, as you suggested.

Actually, I am thinking that maybe it is simpler to store all the SVs in the event without doing any duplicate checks. Because if we want to keep all PF candidates with at least one SV outside of the cone, this necessarily implies that we would need to store quite a lot of SVs anyway (but honestly it's hard for me to tell a priori, if SVs are mostly located outside of the cone or not).

So I view it smth like this: in `TauJet.cc` PF candidates are kept if `hasMatch(pfCand.polarP4()) || SVidx>=0` (so as it is now). No information about associated SV is stored, because actually I realised that in theory one PF candidate can be associated with several SVs. Then outside of this PF loop, `tauJet.secondVertices` are filled with basically all SVs in the event. And in `TauTupleProducer.cc` in `FillSV()` it is like it is now, plus for each SV a vector of PF indices associated to it is stored (in general, without any requirement on tau cone matching etc). But here I am actually not sure, is it possible to implement it like this?

And also I have a concern that this may increase the computation time/tuple size. But I also don't have any intuitive estimate here, so maybe we could try and see how it goes..

msessini:
General question, it seems that EDAnalyzer is now deprecated (??), do you know what should be used instead ?

```warning: 'edm::EDAnalyzer::EDAnalyzer()' is deprecated [-Wdeprecated-declarations]```

**Comment by ofivite** [comment]
> General question, it seems that EDAnalyzer is now deprecated (??), do you know what should be used instead ?

Hmm this is something unusual.. I personally haven't heard about that, so searching for this warning led me to some PRs in CMSSW doing an API change to `edm::one::EDAnalyzer`, for example: https://github.com/cms-sw/cmssw/pull/35343 , https://github.com/cms-sw/cmssw/pull/37044

And then I found this twiki page: https://twiki.cern.ch/twiki/bin/view/CMSPublic/FWMultithreadedFrameworkOneModuleInterface

It looks like they talk about thread safety, but I would need to read this further. @kandrosov do you think it affects us and we would need to migrate too?

File: Production/src/TauJet.cc (hunk 7 above)
Comment (shedprog) [reply Production/src/TauJet.cc]: We need to converge with few things related to the logic of filling SV:
1) what about SVs that match to the TauJet but does not have associated pfCand within this cone, i.e: I propose, to iterate through SVs and fill them is they have PfCand or match cone.
2) Is situation when one pfCand is associated to several SVs is possible. I understand that on the level of matching combinations it is allowed, but will such "fake" SVs be saved in the collection in this case? -> if not, I propose to store index of the SVs for every PfCand and not vice versa.
3) Also, duplicate SVs in the collection would be very bad and would complicate life for the further steps.

I would do it in the following way
```c++

std::set<size_t> matched_sv;

for(size_t pfCandIndex = 0; pfCandIndex < cands_.size(); ++pfCandIndex) {
   ...
   auto SVidx = collectSV(pfCand);
   if(!hasMatch(pfCand.polarP4()) && SVidx<0) continue;
   ...
   if(SVidx >= 0) matched_sv.insert(SVidx);
   // add the index of SV to PfCand and only them push back
   tauJet.cands.push_back(pfCandDesc);
}
...
for(size_t sv_i = 0; sv_i < secondVertices_.size(); ++sv_i) {
   if( !hasMatch(secondVertices_[sv_i].polarP4()) && matched_sv.count(sv_i)) continue;
   out_SVcol.emplace_back(in_SVcol.at(index), index);
   matched_sv.erase(sv_i) 
}

//assert that matched_sv is empty

```
@kandrosov would you agree with such logic?

env.sh
```diff
@@ -153,3 +186,20 @@ else
...
+    Status
+    Docs
+    Contact GitHub
+    Pricing
+    API
+    Training
+    Blog
+    About
+
+TauMLTools/env.sh at master · cms-tau-pog/TauMLTools
```
shedprog: remove please

Production/src/TauJet.cc
shedprog: Sorry, previous solution does not account for correct preserving of indexes of SVs for PfCand for later reference (I am not sure that we need it), but I guess to preserve correct indexes the only option would be putting SVs in the vector before we are putting SV and later putting the rests, probably it is a bit less efficient, but:
```c++
std::map<size_t, size_t> sv_tuple_idxs;

for(size_t pfCandIndex = 0; pfCandIndex < cands_.size(); ++pfCandIndex) {
   ...
   auto SVidx = collectSV(pfCand);
   if(!hasMatch(pfCand.polarP4()) && SVidx<0) continue;
   ...
   pfCandDesc.sv_index = -1; //some default value
   if(SVidx >= 0) {
      if(!sv_tuple_idxs.count(SVidx)) {
         sv_tuple_idxs.insert({SVidx, sv_tuple_idxs.size()};
         tauJet.secondVertices.push_back(secondVertices_.at(SVidx));
      }
      pfCandDesc.sv_index = sv_tuple_idxs.at(SVidx);
   }
   tauJet.cands.push_back(pfCandDesc);
}
...
for(size_t sv_i = 0; sv_i < secondVertices_.size(); ++sv_i) {
   if(matched_sv.count(sv_i)) continue;
   if(!hasMatch(secondVertices_.at(sv_i).polarP4())) continue;
   tauJet.secondVertices.push_back(secondVertices_.at(sv_i));
}
```

Production/plugins/TauTupleProducer.cc
```diff
@@ -1068,23 +1030,52 @@ class TauTupleProducer : public edm::EDAnalyzer {
...
+           	tauTuple().sv_zE.push_back(static_cast<float>(std::sqrt(SV->vertexCovariance(2,2))));
+           	tauTuple().sv_tE.push_back(static_cast<float>(SV->tError()));
+	   	tauTuple().sv_chi2.push_back(static_cast<float>(SV->vertexChi2()));
+           	tauTuple().sv_ndof.push_back(static_cast<float>(SV->vertexNdof()));
+		for(int cand_idx : candsIdx) {
+                    tauTuple().sv_cands_svIdx.push_back(sv_ptr.index);
+                    tauTuple().sv_cands_pfcandIdx.push_back(cand_idx);
+  		}
+	        ++nsv;
+            }
```
shedprog, line 1076: @ofivite correct me if I am wrong, in this way -> SVs that matched to PFCand that selected within cone will be filled only, so then the difference I was thinking to also add the guys that are within the cone but do not have any pfcands. Also, this structure with indexes is a bit weird, I would have indexes for Pfcands pointing SV then it is just <int>

kandrosov: we have chosen SV<->pfCand mapping as a separate vectors, because we were not sure if pfCand->SV is guaranteed to be unique. Do you know if it is unique by design?

kandrosov:
> It looks like they talk about thread safety, but I would need to read this further. @kandrosov do you think it affects us and we would need to migrate too?

it is usually safer to run crab tasks with one thread to reduce memory limit exceptions. I would say it is not urgent to migrate, but we can add it to our todo list to stay aligned with the mainstream CMSSW code.

Production/src/TauJet.cc
```diff
@@ -308,6 +309,25 @@ void TauJetBuilder::Build()
...
+	};
+
+        const auto collectSV = [&](const auto& pfCand) {
+	    auto pfCandPtr = dynamic_cast<const reco::Candidate*>(&pfCand);
+	    for(size_t sv_idx = 0; sv_idx < secondVertices_.size(); ++sv_idx) {
+	        const auto& sv = secondVertices_.at(sv_idx);
+                if(findSVMatch(sv, pfCandPtr))
+		    return static_cast<int>(sv_idx);
+            }
+            return -1;
```
kandrosov, line 328: should we return std::set instead?

kandrosov (changes requested):

Production/src/MuonHitMatch.cc
```diff
@@ -47,7 +47,7 @@ void MuonHitMatch::CountMatches(const pat::Muon& muon, CountMap& n_matches)
 {
     for(const auto& segment : muon.matches()) {
         if(segment.segmentMatches.empty() && segment.rpcMatches.empty() && segment.gemMatches.empty() && segment.me0Matches.empty()) continue;
-        if(n_matches.count(segment.detector())) {
+        if(n_matches.count(segment.detector()) && segment.station()!=0) {
```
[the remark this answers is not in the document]
DennRoy: I don't know if "0" is really valid (My understanding is insufficient to answer this). In order to fix the crash that happens otherwise [here](https://github.com/cms-tau-pog/TauMLTools/blob/master/Production/src/MuonHitMatch.cc#L29), the alternative would be to allow the station 0 [here](https://github.com/cms-tau-pog/TauMLTools/blob/master/Production/interface/MuonHitMatch.h#L16).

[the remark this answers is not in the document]
kandrosov: I agree that the crush must be fixed. But it would be preferable to make a conscious choice whatever we want to ignore stations "0" or it could be useful, and then modify the code accordingly.

Production/src/TauJet.cc
```diff
@@ -364,7 +385,13 @@ void TauJetBuilder::Build()
             tauJet.lostTracks.push_back(pfCandDesc);
         }
 
+        for(size_t svIndex = 0; svIndex < secondVertices_.size(); ++svIndex) {
+   	  const auto& sv = secondVertices_.at(svIndex);
+	  tauJet.secondVertices.emplace_back(sv, static_cast<int>(svIndex));
```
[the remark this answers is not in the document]
msessini: @kandrosov Here I simply add all the SV to the collection so they are later on treated as part of the TauJet::SVCollection. However in '''TauTupleProducer''' only matched SV are stored. Is it fine like this ?

Production/plugins/TauTupleProducer.cc
```diff
@@ -950,6 +1030,54 @@ class TauTupleProducer : public edm::EDAnalyzer {
...
+
+	int nsv = 0;
+        for(const auto& sv_ptr : sv_col) {
+            const reco::VertexCompositePtrCandidate* SV = sv_ptr.obj;
+            auto candsIdx = collectCands(*SV);
+            if(candsIdx.size() > 0){
+	        tauTuple().sv_x.push_back(static_cast<float>(SV->position().x()));
+                tauTuple().sv_y.push_back(static_cast<float>(SV->position().y()));
+           	tauTuple().sv_z.push_back(static_cast<float>(SV->position().z()));
+           	tauTuple().sv_t.push_back(static_cast<float>(SV->t()));
```
shedprog, line 1064: I would also store pt, eta, phi, mass

Production/src/TauJet.cc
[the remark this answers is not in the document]
kandrosov: as it was discussed, it is better to select vertices and pf candidates that associated to tau at this stage. I propose to replace `for(TauJet& tauJet : tauJets_)` as following:
```c++
    for(TauJet& tauJet : tauJets_) {

        auto hasSVPFMatch = [&](const reco::VertexCompositePtrCandidate& sv, const reco::Candidate* pfCandPtr) {
            for(size_t it = 0; it < sv.numberOfSourceCandidatePtrs(); ++it) {
                const edm::Ptr<reco::Candidate>& recoCandPtr = sv.sourceCandidatePtr(it);
                if(pfCandPtr == &(*recoCandPtr))
                    return true;
            }
            return false;
        };

        const auto hasMatch = [&](const PolarLorentzVector& p4) {
            if(tauJet.genLepton && deltaR(p4, tauJet.genLepton->visibleP4()) < setup_.genLepton_cone)
                return true;
            if(tauJet.genJet && deltaR(p4, tauJet.genJet->polarP4()) < setup_.genJet_cone)
                return true;
            if(tauJet.tau && deltaR(p4, tauJet.tau->polarP4()) < setup_.tau_cone)
                return true;
            if(tauJet.boostedTau && deltaR(p4, tauJet.boostedTau->polarP4()) < setup_.boostedTau_cone)
                return true;
            if(tauJet.jet && deltaR(p4, tauJet.jet->polarP4()) < setup_.jet_cone)
                return true;
            if(tauJet.fatJet && deltaR(p4, tauJet.fatJet->polarP4()) < setup_.fatJet_cone)
                return true;
            return false;
        };

        const auto fillMatched = [&](auto& out_col, const auto& in_col) {
            for(size_t index = 0; index < in_col.size(); ++index) {
                const auto& obj = in_col.at(index);
                if(hasMatch(obj.polarP4()))
                    out_col.emplace_back(obj, index);
            }
        };

        for(size_t pfCandIndex = 0; pfCandIndex < lostTracks_.size(); ++pfCandIndex) {
            const auto& pfCand = lostTracks_.at(pfCandIndex);
            if(!hasMatch(pfCand.polarP4())) continue;
            PFCandDesc pfCandDesc;
            pfCandDesc.candidate = &pfCand;
            pfCandDesc.index = static_cast<int>(pfCandIndex);
            tauJet.lostTracks.push_back(pfCandDesc);
        }

        std::map<int, PFCandDesc> selCands;
        std::map<int, const reco::VertexCompositePtrCandidate*> selSVs;
        size_t nSelCands = 0, nSelSVs = 0;
        while(true) {
            for(size_t pfCandIndex = 0; pfCandIndex < cands_.size(); ++pfCandIndex) {
                const auto& pfCand = cands_.at(pfCandIndex);
                if(selCands.count(pfCandIndex)) continue;
                bool belongsToSelectedSV = false;
                for(const auto& [svIndex, sv] : selSVs) {
                    if(hasSVPFMatch(*sv, &pfCand)) {
                        belongsToSelectedSV = true;
                        break;
                    }
                }
                if(!(belongsToSelectedSV || hasMatch(pfCand.polarP4()))) continue;
                PFCandDesc pfCandDesc;
                pfCandDesc.candidate = &pfCand;
                pfCandDesc.index = static_cast<int>(pfCandIndex);
                if(tauJet.tau) {
                    pfCandDesc.tauSignal = IsTauSignalCand(*tauJet.tau, pfCand);
                    pfCandDesc.tauLeadChargedHadrCand = IsLeadChargedHadrCand(*tauJet.tau, pfCand);
                    pfCandDesc.tauIso = IsTauIsoCand(*tauJet.tau, pfCand);
                }
                if(tauJet.boostedTau) {
                    pfCandDesc.boostedTauSignal = IsTauSignalCand(*tauJet.boostedTau, pfCand);
                    pfCandDesc.boostedTauLeadChargedHadrCand = IsLeadChargedHadrCand(*tauJet.boostedTau, pfCand);
                    pfCandDesc.boostedTauIso = IsTauIsoCand(*tauJet.boostedTau, pfCand);
                }
                if(tauJet.jet)
                    pfCandDesc.jetDaughter = IsJetDaughter(*tauJet.jet, pfCand);
                if(tauJet.fatJet)
                    pfCandDesc.subJetDaughter = GetMatchedSubJetIndex(*tauJet.fatJet, pfCand);
                selectedCands[pfCandDesc.index] = pfCandDesc;
            }

            for(size_t svIndex = 0; svIndex < secondVertices_.size(); ++svIndex) {
                const auto& sv = secondVertices_.at(svIndex);
                if(selSVs.count(svIndex)) continue;
                bool hasSelectedPF = false;
                for(const auto& [pfCandIndex, pfCandDesc] : selCands) {
                    if(hasSVPFMatch(sv, pfCandDesc.candidate)) {
                        hasSelectedPF = true;
                        break;
                    }
                }
                if(!(hasSelectedPF || hasMatch(sv.polarP4()))) continue;
                selSVs[static_cast<int>(svIndex)] = &sv;
            }
            if(nSelCands == selCands.size() && nSelSVs == selSVs.size()) break;
            nSelCands = selCands.size();
            nSelSVs == selSVs.size();
        }

        for(const auto& [pfCandIndex, pfCandDesc] : selCands)
            tauJet.cands.push_back(pfCandDesc);
        for(const auto& [svIndex, sv] : selSVs)
            tauJet.secondVertices.emplace_back(sv, svIndex);

        fillMatched(tauJet.electrons, electrons_);
        fillMatched(tauJet.photons, photons_);
        fillMatched(tauJet.muons, muons_);
        fillMatched(tauJet.isoTracks, isoTracks_);
    }
```

Production/plugins/TauTupleProducer.cc
```diff
@@ -950,6 +1030,54 @@ class TauTupleProducer : public edm::EDAnalyzer {
...
+    void FillSV(const TauJet::SVCollection sv_col, const TauJet::PFCandCollection& cands)
+    {
+
+        auto findMatch = [&](const edm::Ptr<reco::Candidate>& recocand) { 
+            for(size_t cand_idx = 0; cand_idx < cands.size(); ++cand_idx) {
+                auto pfcand = dynamic_cast<const reco::Candidate*>(cands.at(cand_idx).candidate);
+                if(pfcand == &(*recocand))
+                   return static_cast<int>(cand_idx);
+            }
+            return -1;
```
kandrosov, line 1042: if TauJet modified as suggested above, no-match situation shouldn't occur. 
```suggestion
            throw std::runtime_error("No SV <-> pfCand match found");
```

kandrosov (changes requested):

Production/plugins/TauTupleProducer.cc
msessini: Do we still actually need this function anyway or filling all SVs from ```selSVs``` is enough ?

kandrosov: we need to compute sv_index <-> pf_index associations. If we run only on SV stored in TauJet, such association should always exist (otherwise there is a bug in SV&pf adding code in TauJet)

Production/plugins/TauTupleProducer.cc
```diff
@@ -950,6 +1001,58 @@ class TauTupleProducer : public edm::EDAnalyzer {
         }
     }
 
+    void FillSV(const TauJet::SVCollection sv_col, const TauJet::PFCandCollection& cands)
```
kandrosov, line 1004:
```suggestion
    void FillSV(const TauJet::SVCollection& sv_col, const TauJet::PFCandCollection& cands)
```

env.sh
```diff
@@ -35,14 +33,19 @@ if [[ $MODE = "prod2018" || $MODE = "phase2" || $MODE = "prod2018UL" || $MODE =
...
         CMSSW_VER=CMSSW_10_6_29
         APPLY_BOOSTED_FIX=0
         export SCRAM_ARCH=slc7_amd64_gcc700
     elif [ $MODE = "run3" ] ; then
-        CMSSW_VER=CMSSW_12_4_0_pre3
+        CMSSW_VER=CMSSW_12_4_10
         APPLY_BOOSTED_FIX=0
-        export SCRAM_ARCH=slc7_amd64_gcc100
+        #export SCRAM_ARCH=el8_amd64_gcc10
+        export SCRAM_ARCH=slc7_amd64_gcc700
```
kandrosov, line 48:
```suggestion
        local os_version=$(cat /etc/os-release | grep VERSION_ID | sed -E 's/VERSION_ID="([0-9]+)"/\1/')
        if [[ $os_version = "7" ]]; then
            export SCRAM_ARCH=slc7_amd64_gcc10
        else
            export SCRAM_ARCH=el8_amd64_gcc10
        fi
```

kandrosov (changes requested):

kandrosov (approved):

ofivite (approved):

shedprog (approved):

PR #129, opened 2022-11-11
Fixed environment and production code for Run2 UL

kandrosov:
- env.sh: added os-based sub-folder for CMSSW installation path
- env.sh: drop support for pre-ul 2018 and boosted tau fix
- Production.py: way to access GT is the same for run2 UL and run3


Production/python/Production.py
```diff
@@ -92,7 +92,7 @@
     process.load('Configuration.StandardSequences.FrontierConditions_GlobalTag_cff')
```
ofivite, line 92: Few things/questions  which I've just noticed, not fully related to the changes:
- Would it makes sense to add v2.5 for `isRun2UL` case similarly to `isRun3`? I see that in `env.sh` they both are set up with 12.4.X  so v2.5 should be runnable in this mode too https://github.com/cms-tau-pog/TauMLTools/blob/04a6aab8b160a083d5b5c50686d8e120fc60dee7/Production/python/Production.py#L148-L158
- I think we can drop now `isRun2PreUL` flag ([L52](https://github.com/cms-tau-pog/TauMLTools/blob/04a6aab8b160a083d5b5c50686d8e120fc60dee7/Production/python/Production.py#L52), [L319](https://github.com/cms-tau-pog/TauMLTools/blob/04a6aab8b160a083d5b5c50686d8e120fc60dee7/Production/python/Production.py#L139))
- I am not sure that these lines are needed, the Path for `isPhase2` already includes `slimmedElectronsMerged` right?: https://github.com/cms-tau-pog/TauMLTools/blob/04a6aab8b160a083d5b5c50686d8e120fc60dee7/Production/python/Production.py#L316-L317
- About the `else` statements in [L159](https://github.com/cms-tau-pog/TauMLTools/blob/04a6aab8b160a083d5b5c50686d8e120fc60dee7/Production/python/Production.py#L159) and [L295](https://github.com/cms-tau-pog/TauMLTools/blob/04a6aab8b160a083d5b5c50686d8e120fc60dee7/Production/python/Production.py#L295): does it basically target now only embedded samples and data (since we drop preUL support)? Would it make sense to make it explicit and raise exception otherwise? 
- Is the `boostedSequence` needed now or having a common for all eras `boostedTaus_InputTag = cms.InputTag('slimmedTausBoosted')` is ok?

TauMLTools.code-workspace
```diff
@@ -0,0 +1,8 @@
+{
```
ofivite, line 1: Looks like this file was accidentally committed, please remove it.

ofivite (changes requested):

Production/python/Production.py
kandrosov: done

TauMLTools.code-workspace
kandrosov: I thought it would be nice having the project file in the repository so one doesn't need to re-create it each time. But if you prefer, I can remove it.

ofivite: Oh sorry for my ignorance, when commenting I thought that it's just some internally used vscode (?) file. What would be its use-case scenario?

kandrosov: usually I edit code with VSCode using remote ssh plugin and it is more convenient to have a workspace file. But it is also not a big deal to re-create it for each new installation. So up to you.

ofivite: That's interesting because I also use vscode+ssh remote but I've never set up anything like this (so ssh tunnels worked out of the box for me). But anyhow it's not a problem to keep the file.

kandrosov: Yes it works, but sometimes it is nice to have project-dependent settings. For example, one silly use-case can be: set in `TauMLTools.code-workspace`
```
	"settings": {
		"editor.tabSize": 2
	}
```
to unify tabulation size within TauMLTools (assuming everyone is using vscode)

ofivite (changes requested):

TauMLTools.code-workspace
ofivite: Nice, good to learn about that, thanks! I am personally not familiar with the syntax to fill it now so that we have a unified code style across the group. So let's keep this file as it is now and changes can be introduced in the future.

ofivite (approved):

shedprog (approved):

PR #130, opened 2022-11-15
Customise HLT for Deeptau 

zhiyuanlcern:


zhiyuanlcern:
Hi Konstantin and others,
For testing I am using this file: /store/mc/Run3Summer21DRPremix/TT_TuneCP5_14TeV-powheg-pythia8/GEN-SIM-DIGI-RAW/120X_mcRun3_2021_realistic_v6-v2/2540000/00c642d1-bf7e-477d-91e1-4dd9ce2c8099.root

I added a few variables that are directly accessible at the reco level (I used https://github.com/cms-sw/cmssw/tree/master/DataFormats/Candidate/interface/Candidate.h as reference), the code ran through and produced a nano.root and DQMIO.root file.
While checking these new variable distributions, I noticed that distributions like et, energy, mt etc look ok. But variables like isMuon, isElectron, isJet, vertexNdof etc give 0. I wonder why is that?
Another question, could you point me to where these variables are actually calculated? I checked Candidate.cc and PFCandidate.cc, the former is more just an abstract of variables to be defined and later seems to be doing something else. I imagine there would be somewhere where these variables are actually calculated.

Cheers,
Jordan

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,159 @@
...
+      et2 = Var("et2",float, doc= 'transverse energy squared (use this for cut!)'),
+      massSqr = Var("massSqr",float, doc= 'mass squared'),
+      mt = Var("mt",float, doc= 'transverse mass'),
+      mtSqr = Var("mtSqr",float, doc= 'transverse mass squared'),
+      px = Var("px",float, doc= 'x coordinate of momentum vector'),
+      py = Var("py",float, doc= 'y coordinate of momentum vector'),
+      pz = Var("pz",float, doc= 'z coordinate of momentum vector'),
+      theta = Var("theta",float, doc= 'momentum polar angle'),
+      rapidity = Var("rapidity", float, doc='rapidity'),
+      y = Var("y", float, doc='rapidity'),
```
kandrosov, line 97:
```suggestion
```
all these variables can be re-computed based on (pt, eta, phi, mass) stored above => no need to save them

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,159 @@
...
+      vx = Var("vx", float, doc='x coordinate of vertex position'),
+      vy = Var("vy", float, doc='y coordinate of vertex position'),
+      vz = Var("vz", float, doc='z coordinate of vertex position'),
+      vertexChi2 = Var("vertexChi2", float, doc='chi-squares'),
+      vertexNdof = Var("vertexNdof", float, doc='Number of degrees of freedom,  Meant to be Double32_t for soft-assignment fitters'),
+      vertexNormalizedChi2 = Var("vertexNormalizedChi2", float, doc='chi-squared divided by n.d.o.f.'),
```
kandrosov, line 103:
```suggestion
```
can be recomputed using vertexChi2/vertexNdof

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,159 @@
...
+      hasMasterClonePtr = Var("hasMasterClonePtr", bool, doc='This only happens if the concrete Candidate type is ShallowClonePtrCandidate'),
+      isElectron = Var("isElectron", bool, doc=''),
+      isMuon = Var("isMuon", bool, doc=''),
+      isStandAloneMuon = Var("isStandAloneMuon", bool, doc=''),
+      isGlobalMuon = Var("isGlobalMuon", bool, doc=''),
+      isTrackerMuon = Var("isTrackerMuon", bool, doc=''),
+      isCaloMuon = Var("isCaloMuon", bool, doc=''),
+      isPhoton = Var("isPhoton", bool, doc=''),
+      isConvertedPhoton = Var("isConvertedPhoton", bool, doc=''),
+      isJet = Var("isJet", bool, doc=''),
```
kandrosov, line 118: these variables will add dependency on reconstruction of other object types (e.g. reco electron, muon etc.). I suggest to not use them for this round and rely only on `pdgId` that is already stored above
```suggestion
```

kandrosov (changes requested):
> Another question, could you point me to where these variables are actually calculated?

depending on variable it could be computed in different places. Since we not going to use `isElectron` etc. (see other comment), it is not urgent to search their definitions in the code. Please, go ahead with implementation of track- and calo-related variables for pf candidates.

Production/python/customiseHLT.py
zhiyuanlcern: Ok, ty for the info. But for muon there seem to be different types of it -- is this info not needed neither?

kandrosov: Muon faking tau is usually not a big issue at HLT level. Perhaps, it is relevant for mu+tau cross-trigger, but even there, simple cut-based selection is enough to reduce Z->mumu background. Therefore, for online tagging we should more focus on jets and electrons (with lower priority) faking taus.

zhiyuanlcern: Ok, thanks a lot for your explanation!

zhiyuanlcern:
Another question (a more general one), why are we using variables at reco level? Shouldn'w we look at the RAW level data?

Production/python/customiseHLT.py
zhiyuanlcern: Another question (a more general one), why are we using variables at reco level? Shouldn'w we look at the RAW level data?

kandrosov: At L3, HLT uses the same code base as RECO, so for our use-cases the definition of variables for taus and PF candidates will be the same (from the code perspective) for HLT and offline reco. But we will need to add also some HLT specific collections (e.g. pixel tracks, L1 seeds) that could be useful for online trainings.

Production/python/customiseHLT.py
```diff
@@ -92,8 +92,20 @@ def customise(process):
...
+      CKFtrackDxy = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.dxy : -999.", float, doc = "CKF track Dxy"),
+      GSFtrackDxy = Var("? gsfTrackRef.isNonnull && gsfTrackRef.isAvailable ? gsfTrackRef.dxy : -999.", float, doc = "GSF track Dxy"),
+      CKFtrackIsValid = Var("trackRef.isNonnull && trackRef.isAvailable", bool, doc = "CKF track is valid"),
+      GSFtrackIsValid = Var("gsfTrackRef.isNonnull && gsfTrackRef.isAvailable", bool, doc = "GSF track is valid"),
+      
+      # CKFLostInnerHits = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.hitPattern.numberOfLostHits(reco::HitPattern::MISSING_INNER_HITS) : -999.", float, doc = "CKF track LostInnerHit"), 
+      # This doesn't work: failed to parse "? trackRef.isNonnull && trackRef.isAvailable ? trackRef.hitPattern.numberOfLostHits(reco::HitPattern::MISSING_INNER_HITS) : -999."
+      # also tried: CKFLostInnerHits = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.hitPattern.numberOfLostHits : -999.", float, doc = "CKF track LostInnerHit"), 
+      # this doesn't work: method named "numberOfLostHits" for type "reco::HitPattern" was passed the wrong number of arguments.
```
kandrosov, line 106: I think it is ok to not have these variables for the current round. But I suggest adding following variables (in addition to what you have already implemented):
```
track_dxyError
track_dzError
track_pt
track_eta
track_phi
track_ptError
track_etaError
track_phiError
track_chi2
track_ndof

track_numberOfValidHits
track_numberOfLostHits
track_missingInnerHits
track_missingOuterHits
track_validFraction
track_qualityMask

hcalEnergy
ecalEnergy
```

kandrosov (changes requested):

Production/python/customiseHLT.py
```diff
@@ -92,8 +92,20 @@ def customise(process):
...
       massConstraint = Var("massConstraint", bool, doc='do mass constraint?'),
-
-
+      # source: cmssw/PhysicsTools/NanoAOD/plugins/SimpleFlatTableProducerPlugins.cc 
+      CKFtrackDz = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.dz : -999.", float, doc = "CKF track Dz"),
+      GSFtrackDz = Var("? gsfTrackRef.isNonnull && gsfTrackRef.isAvailable ? gsfTrackRef.dz : -999.", float, doc = "GSF track Dz"),
+      CKFtrackDxy = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.dxy : -999.", float, doc = "CKF track Dxy"),
+      GSFtrackDxy = Var("? gsfTrackRef.isNonnull && gsfTrackRef.isAvailable ? gsfTrackRef.dxy : -999.", float, doc = "GSF track Dxy"),
+      CKFtrackIsValid = Var("trackRef.isNonnull && trackRef.isAvailable", bool, doc = "CKF track is valid"),
+      GSFtrackIsValid = Var("gsfTrackRef.isNonnull && gsfTrackRef.isAvailable", bool, doc = "GSF track is valid"),
```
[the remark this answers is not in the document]
zhiyuanlcern: A comment on the GSF track variables, I checked with more stats (5k events) and plotted GSFtrackDz, still got 0 for all events.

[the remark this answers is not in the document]
kandrosov: thanks for the check! Indeed, after closer look into the sequence, there are no GSF tracks in `HLTParticleFlowSequenceForTaus` => we can drop GSF tracks as well as "CKF" prefix for normal tracks in var names (so variables will be `track_dz` etc.)

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,147 @@
...
+  process.nanoTableTaskFS = cms.Task(process.genParticleTablesTask, process.genParticleTask,
+                                     process.tauTablesTask, process.pfCandTablesTask)
+  process.nanoSequenceMC = cms.Sequence(process.nanoTableTaskFS)
+  process.finalGenParticles.src = cms.InputTag("genParticles")
+
+
+  process.MessageLogger.cerr.FwkReport.reportEvery = 100
+  process = customiseGenParticles(process)
+
+  process.genParticleTable.variables.vx = Var('vertex().x', float, precision=10, doc='x coordinate of the gen particle production vertex')
```
ofivite, line 124: It looks like you already fill these genvertex variables in `customiseGenParticles()`, so these lines can be removed as duplicates.

ofivite (changes requested):

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,147 @@
...
+import FWCore.ParameterSet.Config as cms
+from PhysicsTools.NanoAOD.common_cff import Var, P4Vars
+
+def customiseGenParticles(process):
+  def pdgOR(pdgs):
+    abs_pdgs = [ f'abs(pdgId) == {pdg}' for pdg in pdgs ]
+    return '( ' + ' || '.join(abs_pdgs) + ' )'
+
+  leptons = pdgOR([ 11, 13, 15 ])
+  important_particles = pdgOR([ 6, 23, 24, 25, 35, 39, 9990012, 9900012 ])
```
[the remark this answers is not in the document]
zhiyuanlcern: @kandrosov Could you comment on this question?

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,187 @@
...
+      trackIsValid = Var("trackRef.isNonnull && trackRef.isAvailable", bool, doc = "track is valid"),
+      
+      trackDzError = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.dzError : -999.", float, doc = "track DzError"),
+      trackDxyError = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.dxyError : -999.", float, doc = "track DxyError"),
+      
+      trackPt = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.pt : -999.", float, doc = "track Pt"),
+      trackEta = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.eta : -999.", float, doc = "track Eta"),
+      trackPhi = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.phi : -999.", float, doc = "track Phi"),
+      trackPtError = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.ptError : -999.", float, doc = "track PtError"),
+      trackPtaError = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.etaError : -999.", float, doc = "track PtaError"),
```
ofivite, line 107: small typo, should be `trackEtaError`
```suggestion
      trackEtaError = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.etaError : -999.", float, doc = "track PtaError"),
```

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,187 @@
...
+      trackChi2 = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.chi2 : -999.", float, doc = "track Chi2"),
+      trackNdof = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.ndof : -999.", float, doc = "track Ndof"),
+      
+      # source: DataFormats/TrackReco/interface/HitPattern.h
+      
+      trackNumberOfValidHits = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.hitPattern.numberOfValidHits : -999.", float, doc = "number of valid hits found"),
+      trackNumberOfLostHits = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.numberOfLostHits : -999.", float, doc = " number of cases where track crossed a layer without getting a hit."),
+      trackMissingInnerHits = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.missingInnerHits : -999.", float, doc = "number of hits expected from inner track extrapolation but missing"),
+      missingOuterHits= Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.missingOuterHits : -999.", float, doc = " number of hits expected from outer track extrapolation but missing"),
```
ofivite, line 116: for the sake of consistency in naming:
```suggestion
      trackMissingOuterHits= Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.missingOuterHits : -999.", float, doc = " number of hits expected from outer track extrapolation but missing"),
```

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,187 @@
...
+      validFraction= Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.validFraction : -999.", float, doc = "fraction of valid hits on the track"),
```
ofivite, line 117: for the sake of consistency in naming:
```suggestion
      trackHitsValidFraction= Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.validFraction : -999.", float, doc = "fraction of valid hits on the track"),
```

ofivite (changes requested):

Production/python/customiseHLT.py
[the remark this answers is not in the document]
kandrosov: - 9990012/9900012 are HNL (Majorana and Dirac). We'll use HNL samples to train on displaced taus. Perhaps we can also add stau to the list. @shedprog what pdg ids staus have in our MC?
- genJets will be stored as a separate collection equivalent to the `GenJet` nano collection. So if tau doesn't match to any lepton, but matches to genJet - it will be labeled as jet faking tau.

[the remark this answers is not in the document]
shedprog: In the maximally mixed scenario that we are considering, it is +1000015/-1000015, in other scenarios it might be the number with 7 numbers always ends on *15

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,147 @@
...
+  process.NANOAODSIMoutput_step = cms.EndPath(process.NANOAODSIMoutput)
+
+  process.tauTable = cms.EDProducer("SimpleCandidateFlatTableProducer",
+    src = cms.InputTag( "hltHpsPFTauProducer" ),
+    cut = cms.string(""),
+    name= cms.string("Tau"),
+    doc = cms.string("HLT taus"),
+    singleton = cms.bool(False), # the number of entries is variable
+    extension = cms.bool(False), # this is the main table
+    variables = cms.PSet(
```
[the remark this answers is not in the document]
kandrosov: yes, these are HPS taus. And yes, more variables should be added. Currently @zhiyuanlcern was focusing on pfcand variables. The content of the following collections still needs to be defined are:
- [ ] HPS taus
- [ ] L1 objects: electrons, muons, taus and jets
- [ ] pixel tracks
- [ ] calo rechits

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,158 @@
...
+      P4Vars,
+      charge = Var("charge", int, doc="electric charge"),
+    )
+  )
+  # cms.EDProducer : an object that produces a new data object
+
+  process.pfCandTable = cms.EDProducer("SimpleCandidateFlatTableProducer",
+    src = cms.InputTag( "hltParticleFlowForTaus" ),
```
shedprog, line 76: I wanted to ask: does this collection contain only pfCands that are part of the HLT TAU, or all pfCands in cone? i.e: would you need to add the flags that this is the part HLT TAU object?

kandrosov: it contains **all** pf candidates in the event. For the training we can select all pf cands with `deltaR < X` HPS tau or AK4 jet (if we decide to not use HPS) or L1 tau seed

Production/python/customiseHLT.py
```diff
@@ -6,6 +6,12 @@
 # file dataset=/store/mc/Run3Summer21DRPremix/TT_TuneCP5_14TeV-powheg-pythia8/GEN-SIM-DIGI-RAW/120X_mcRun3_2021_realistic_v6-v2/2540000/00c642d1-bf7e-477d-91e1-4dd9ce2c8099.root
 import FWCore.ParameterSet.Config as cms
 from PhysicsTools.NanoAOD.common_cff import Var, P4Vars
+from RecoJets.JetProducers.ak4GenJets_cfi import ak4GenJets
+from PhysicsTools.NanoAOD.jetMC_cff import Var, P4Vars
```
ofivite, line 10: Wouldn't it be a conflict with `Var/P4Vars` imported earlier? I would assume that these are two different entities

Production/python/customiseHLT.py
```diff
@@ -59,7 +66,61 @@ def customise(process):
...
+        #  1000012, 1000014, 1000016,
+        #  2000012, 2000014, 2000016,
+        #  1000039, 5100039,
+        #  4000012, 4000014, 4000016,
+        #  9900012, 9900014, 9900016,
+        #  39, 
+         12,14,16 ), # ignore for neutrinos
+    partonicFinalState = cms.bool(False),
+    excludeResonances = cms.bool(False),
+    excludeFromResonancePids = cms.vuint32(12, 13, 14, 16), # why 13 ????
```
ofivite: Indeed why:) by logic I'd say it should be only 12,14,16.

ofivite (changes requested):

Production/python/customiseHLT.py
```diff
@@ -71,48 +73,80 @@ def customise(process):
...
+      # PFCand_vertexNdof
+      # PFCand_longLived = Var("longLived", bool, doc='is long lived?'),
+      # PFCand_trackMissingInnerHits = Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.missingInnerHits : -999.", float, doc = "number of hits expected from inner track extrapolation but missing"),
+      # PFCand_trackMissingOuterHits= Var("? trackRef.isNonnull && trackRef.isAvailable ? trackRef.missingOuterHits : -999.", float, doc = " number of hits expected from outer track extrapolation but missing"),
+      # PFCand_status = Var("status", int, doc='status word'),
+      # PFCand_longLived = Var("longLived", bool, doc='is long lived?'),
+      # PFCand_massConstraint = Var("massConstraint", bool, doc='do mass constraint?'),
+
       pdgId = Var("pdgId", int, doc='PDG identifier'),
+      dz = Var("? leadPFCand.trackRef.isNonnull && leadPFCand.trackRef.isAvailable ? leadPFCand.trackRef.dz : -999 ", float, doc='lead PF Candidate dz'),
```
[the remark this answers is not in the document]
zhiyuanlcern: The dxy and its error can't be reached directly here, instead, they are retrieved using an external c++ code (see plugin/TauTableProducerHLT.cc)

Production/python/customiseHLT.py
zhiyuanlcern: This duplication seems to be added by mistake, thanks for pointing it out! I will remove the duplicate import.

ofivite:
@kandrosov For my understanding, unless I'm mistaken you were planning to add the remaining collections (pixel tracks, calor rechits, L1 objects). Will it go also into this PR or in a separate one?

kandrosov:
> @kandrosov For my understanding, unless I'm mistaken you were planning to add the remaining collections (pixel tracks, calor rechits, L1 objects). Will it go also into this PR or in a separate one?

I think at this point it would be better to have a separate PR, if you agree.

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,327 @@
...
+    excludeResonances = cms.bool(False),
+    excludeFromResonancePids = cms.vuint32(12, 14, 16), 
+    tausAsJets = cms.bool(False)
+)
+  process.GenJetTable = cms.EDProducer("SimpleCandidateFlatTableProducer",
+    src = cms.InputTag( "ak4GenJetsNoNu" ),
+    cut = cms.string(""),
+    name= cms.string("GenJet"),
+    doc = cms.string("HLT ak4 GenJet"),
+    SkipEvent = cms.untracked.vstring('ProductNotFound'),
```
kandrosov, line 102:
```suggestion
```

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,327 @@
...
+    + process.HLTGlobalPFTauHPSSequence \
+    + process.HLTHPSDeepTauPFTauSequenceForVBFIsoTau \
+    + process.nanoSequenceMC)
+  process.NANOAODSIMoutput_step = cms.EndPath(process.NANOAODSIMoutput)
+  
+  from PhysicsTools.JetMCAlgos.HadronAndPartonSelector_cfi import selectedHadronsAndPartons
+  process.selectedHadronsAndPartons = selectedHadronsAndPartons.clone(src = cms.InputTag("genParticlesForJetsNoNu"))
+  # set ak4GenJets producer 
+  # process.ak4GenJetsNoNu = ak4GenJets.clone( src = "genParticlesForJetsNoNu")
+  process.ak4GenJetsNoNu = ak4GenJets.clone( src = "genParticlesForJetsNoNu")#, bHadrons= cms.InputTag("selectedHadronsAndPartons","bHadrons"), cHadrons= cms.InputTag("selectedHadronsAndPartons","cHadrons"),partons= cms.InputTag("selectedHadronsAndPartons","physicsPartons"), )
```
kandrosov, line 74:
```suggestion
  process.ak4GenJetsNoNu = ak4GenJets.clone( src = "genParticlesForJetsNoNu")
```

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,327 @@
...
+        #  1000039, 5100039,
+        #  4000012, 4000014, 4000016,
+        #  9900012, 9900014, 9900016,
+        #  39, 
+         12,14,16 ), # ignore for neutrinos
+    partonicFinalState = cms.bool(False),
+    excludeResonances = cms.bool(False),
+    excludeFromResonancePids = cms.vuint32(12, 14, 16), 
+    tausAsJets = cms.bool(False)
+)
```
kandrosov, line 96:
```suggestion
  process.load('RecoJets.Configuration.GenParticles_cff')
```

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+
+def customiseGenParticles(process):
+  def pdgOR(pdgs):
+    abs_pdgs = [ f'abs(pdgId) == {pdg}' for pdg in pdgs ]
+    return '( ' + ' || '.join(abs_pdgs) + ' )'
+
+  leptons = pdgOR([ 11, 13, 15 ])
+  important_particles = pdgOR([ 6, 23, 24, 25, 35, 39, 9990012, 9900012 ])
```
kandrosov, line 22:
```suggestion
  important_particles = pdgOR([ 6, 23, 24, 25, 35, 39, 9990012, 9900012, 1000015 ])
```

Production/plugins/GenJetFlavourTableProducerHLT.cc
```diff
@@ -0,0 +1,95 @@
...
+#include "DataFormats/JetMatching/interface/JetFlavourInfo.h"
+#include "DataFormats/JetMatching/interface/JetFlavourInfoMatching.h"
+#include "DataFormats/Math/interface/deltaR.h"
+
+#include "DataFormats/NanoAOD/interface/FlatTable.h"
+
+#include "CommonTools/Utils/interface/StringCutObjectSelector.h"
+#include "CommonTools/Utils/interface/StringObjectFunction.h"
+
+class GenJetFlavourTableProducerHLT : public edm::stream::EDProducer<> {
```
kandrosov, line 24: doesn't need to duplicate the code. [GenJetFlavourTableProducer](https://github.com/cms-sw/cmssw/blob/master/PhysicsTools/NanoAOD/plugins/GenJetFlavourTableProducer.cc) from CMSSW should work fine for our case

Production/python/customiseHLT.py
```diff
@@ -0,0 +1,327 @@
...
+      rawHcalEnergy = Var("rawHcalEnergy", float, doc='rawHcalEnergy'),
+      rawEcalEnergy = Var("rawEcalEnergy", float, doc='rawEcalEnergy'),
+      EcalEnergy = Var("ecalEnergy", float, doc='EcalEnergy'),
+      HcalEnergy = Var("hcalEnergy", float, doc='HcalEnergy'),
+    )
+  )
+  process.AK4PFJetsTable = cms.EDProducer("SimpleCandidateFlatTableProducer",
+    src = cms.InputTag( "hltAK4PFJets" ),
+    cut = cms.string(""),
+    name= cms.string("AK4PFJets"),
```
kandrosov, line 234:
```suggestion
    name= cms.string("Jet"),
```
(to be compatible with standard nano notation)

kandrosov (changes requested):

Production/python/customiseHLT.py
zhiyuanlcern: I am looking at https://github.com/cms-sw/cmssw/blob/master/RecoJets/Configuration/python/GenJetParticles_cff.py.

This should be GenJetParticles_cff not GenParticles_cff? 

In addition, I see in the official repo there is `excludeFromResonancePids = cms.vuint32(12, 13, 14, 16),` @ofivite suggested removing 13 IIRC. Do you know if 13 should be on the list?

kandrosov: > This should be GenJetParticles_cff not GenParticles_cff?

Indeed, it should be `GenJetParticles_cff`



> In addition, I see in the official repo there is `excludeFromResonancePids = cms.vuint32(12, 13, 14, 16),` @ofivite suggested removing 13 IIRC. Do you know if 13 should be on the list?

In that config two producers are defined: `genParticlesForJetsNoNu` and `genParticlesForJetsNoMuNoNu`. Muons are excluded in the second one, while for our purposes `genParticlesForJetsNoNu` should be used.

kandrosov: And `excludeResonances`is set to `False`, so what is specified in `excludeFromResonancePids` is not relevant.

zhiyuanlcern: Ah ok, that answers my questions. Thanks!

kandrosov (approved):

ofivite (approved):

PR #131, opened 2022-11-24
Fix of the scaling vector filling bug

shedprog:


Training/interface/DataLoader_main.h
```diff
@@ -1,3 +1,4 @@
+#include "TauMLTools/Analysis/interface/TauTuple.h"
```
kandrosov, line 1: `TauTuple` code is generated automatically during loading => this include should be removed

Training/python/config_parse.py
```diff
@@ -177,9 +177,16 @@ def create_scaling(content_scaling: dict, content_cfg: dict) -> str:
                 string += "std::vector<std::vector<float>> "
                 string += subg + " = "
                 var_string = []
-                for var_i, (var, var_params) in enumerate(content_scaling[FeatureT].items()):
-                    assert var in content_cfg['Features_all'][FeatureT][var_i].keys() # check if there is such feature in training cfg
-                    if var in content_cfg['Features_disable'][FeatureT]: continue
+
```
ofivite, line 180: For the sake of consistency I would also change the iteration over the feature groups based on the training config (instead of the scaling cfg currently):
https://github.com/cms-tau-pog/TauMLTools/blob/f6342fe68bc080731bcf6f84193cdd19011770cc/Training/python/config_parse.py#L172

Training/python/config_parse.py
```diff
@@ -177,9 +177,16 @@ def create_scaling(content_scaling: dict, content_cfg: dict) -> str:
...
+                feature_list_enabled= []
+                for feature_dict in content_cfg["Features_all"][FeatureT]:
+                    assert len(feature_dict) == 1 and type(feature_dict) == dict
+                    if list(feature_dict)[0] in content_cfg["Features_disable"][FeatureT]:
+                        continue
+                    feature_list_enabled.append(list(feature_dict)[0])
+
+                for var_i, var in enumerate(feature_list_enabled):
```
ofivite, line 188: [minor nitpicking] `var_i` from enumeration isn't used so it can be removed

ofivite (changes requested):

Training/interface/DataLoader_main.h
shedprog: ok, then let's then substitute the testing script with the ability to test only DataLoader all together (not only c++ part)

ofivite: may I ask just for my understanding, why this include is needed in general?

shedprog: during the development, it was easy to have several module testing steps, one of the script (DataLoader_test.py) was testing only c++ part of the dataloader i.e: before all the functionality that is included in `class DataLoader`, in this testing script compilation of the classes was not included, so I added this, but indeed it may conflict with `makeTree` functionality on the later steps of using dataloader. Now one should use `python DataLoader_fulltest.py` for testing

ofivite (approved):

PR #132, opened 2022-11-28
Gen lepton with > 6 mothers

msessini:
This PR includes a warning for gen lepton with more than 6 mothers and store only the 6 firsts.

The cms::Exception is kept for safety but shouldn't occur anymore.


Analysis/interface/GenLepton.h
```diff
@@ -243,8 +243,15 @@ class GenLepton {
             size_t last_mother_index = NoneIndex;
 
             if(!particle->motherRefVector().empty()) {
-                for(const auto& mother : particle->motherRefVector())
+		int Nmother = 0;
```
kandrosov, line 246: I think it is better to fix the limitation at the level of filling TauTuple and keep `GenLepton.h` logic intact.
https://github.com/cms-tau-pog/TauMLTools/blob/master/Production/plugins/TauTupleProducer.cc#L401-L420

Analysis/interface/GenLepton.h
```diff
@@ -243,8 +243,15 @@ class GenLepton {
...
+                for(const auto& mother : particle->motherRefVector()) {
                     FillDaughters(mother.get(), NoneIndex, false);
+		    ++Nmother;
+		    if(Nmother>6) {
+			    std::cout<<"Warning : Gen particle with > 6 mothers, storing only 6 firsts"<<std::endl;
+			    break;
+		    }
```
shedprog, line 253: May I ask: in which MC it is observed? Is it for pp interaction level summing up over all the partons? I am a bit concerned why do we want to save it

kandrosov: Having more than 2 mothers occur regularly in all our MC in tau decays from FSR. The total number of mothers in fragmentation does not have an upper limit (from physics perspective). For technical reasons, we can save only up to 6 mothers. But we don't need to have the precise decay graph for FSR products, so I think it is ok to store only first 6 mothers.

Analysis/interface/GenLepton.h
```diff
@@ -243,6 +243,7 @@ class GenLepton {
             size_t last_mother_index = NoneIndex;
 
             if(!particle->motherRefVector().empty()) {
+		int Nmother = 0;
```
kandrosov, line 246: `Nmother` is not used -> remove

Production/plugins/TauTupleProducer.cc
```diff
@@ -420,6 +425,8 @@ class TauTupleProducer : public edm::EDAnalyzer {
             };
 
             tauTuple().genLepton_lastMotherIndex = static_cast<int>(genLepton->mothers().size()) - 1;
+	    if(genLepton->mothers().size()>6) 
+		    tauTuple().genLepton_lastMotherIndex = 5;
```
kandrosov, line 429:
```suggestion
            tauTuple().genLepton_lastMotherIndex = static_cast<int>(std::min(genLepton->mothers().size(), 6)) - 1;
```

Production/plugins/TauTupleProducer.cc
```diff
@@ -403,15 +403,20 @@ class TauTupleProducer : public edm::EDAnalyzer {
...
                 Long64_t pos = 0;
                 Long64_t shift = 1;
                 std::set<int> mother_indices;
-                for(auto mother : mothers)
+		int Nmother = 0;
+                for(auto mother : mothers) {
+		    if(Nmother>6) break;
                     mother_indices.insert(getIndex(mother));
+		    ++Nmother;
+		}
```
kandrosov, line 419:
```suggestion
                for(auto mother : mothers) {
		    if(mother_indices.size() >= 6) break;
                    mother_indices.insert(getIndex(mother));
		}
```

kandrosov (changes requested):

kandrosov (approved):

ofivite (approved):

shedprog (approved):

PR #133, opened 2022-12-01
TF training workflow

ofivite:
This PR merges developments done in a separate [framework](https://github.com/ofivite/tatoo), adds some minor refactoring w.r.t. to the main branch status https://github.com/ofivite/tatoo/commit/4a97d728a11e8ec3eb36c3e0c1dbe4b669c0f92d and removes some unused files from the training directory. 

In general, the parameters values in the config files correspond to the latest transformer training I did. A brief overview of the introduced training workflow can be found [here](https://indico.cern.ch/event/1209780/#2-tatoo-framework), and in a nutshell update adds the following steps:

**ROOT -> tf.data preprocessing**. This is done with [`create_dataset.py`](https://github.com/ofivite/TauMLTools/blob/tf-pipe/Training/tf/create_dataset.py) script. The main config file is [`create_dataset.yaml`](https://github.com/ofivite/TauMLTools/blob/tf-pipe/Training/tf/configs/create_dataset.yaml) and it merges three more configs from `configs/data`: `features.yaml`, `input_samples.yaml`, `gen_config.yaml`. After specifying the data files to be run on in the former cfg file it can be run with  (it's also possible to override params in CLI):
```bash
python create_dataset.py dataset_name=alias_to_be_used_later path_to_dataset=directory_where_to_save_tf_skims
```
This will produce for each of the ROOT files (specified in `files` key of `input_samples.yaml`) a `tf.dataset` of `tf.RaggedTensor` separately for each of the tau types (specified as a list next to the file name entry). 

**Training**. This step is done with [`train.py`](https://github.com/ofivite/TauMLTools/blob/tf-pipe/Training/tf/train.py) with the main config being [`train.yaml`](https://github.com/ofivite/TauMLTools/blob/tf-pipe/Training/tf/configs/train.yaml). The latter also imports configurations from `configs/data/features.yaml` and `configs/model/model_to_be_used.yaml`. After specifying the dataset names (defined at the previous step) and other necessary configuration for the training one can run it with:
```bash
python train.py experiment_name=my_experiment model=model_name n_epochs=1 # model_name={transformer, particle_net}
```
N.B.: `particle_net` and `transformer` models should work out-of-the box in this setup. The `taco` interface (custom implementation of this [paper](https://arxiv.org/abs/2107.02908)) is deprecated (in terms of embedding/decoder layer + cfgs) but the core should be functional. One can therefore use it for future R&D. 

This step will train the model on the mixture of datasets and tau types (obtained at the previous step) as defined in the input configuration. A uniform sampling of objects across the target classes is performed. The outputs are logged to mlflow as usual. 

**Prediction**. This is a workflow similar to that we have now, only defined on the preprocessed `tf.dataset` files. It has the main cfg [`configs/predict.yaml`](https://github.com/ofivite/TauMLTools/blob/tf-pipe/Training/tf/configs/predict.yaml) and  can be run as:
```bash
python predict.py # assuming all the params are specified in the `predict.yaml` cfg
```
The predictions are logged to mlflow artifacts and one can use them in the usual evaluation workflow (although some changes need to be introduced, to be checked)

**Attention visualisation** as developed by @ccmfb and presented [here](https://indico.cern.ch/event/1197016/#3-status-of-deeptau-v3).  The core modules are kept in [utils/visualize.py](https://github.com/cms-tau-pog/TauMLTools/compare/master...ofivite:TauMLTools:tf-pipe?expand=1#diff-207e3d7ed468e9b7cbd8facea8f9fb4935c536d09ada455bb90ba6ebd4cb14b4) and interactive widgets can be created in the notebook [visualization_notebook.ipynb](https://github.com/cms-tau-pog/TauMLTools/compare/master...ofivite:TauMLTools:tf-pipe?expand=1#diff-54f7f122fa29658fcb9f57f8c3665dab3bc2f5cc02fe2e641dea6d6b65ba26dd). This part can also be used for future R&D.


kandrosov (changes requested):
I have tried to run on lxplus with current TauML env, and I'm getting this error:
```
python Training/tf/create_dataset.py dataset_name=DY path_to_dataset=output/DY
Traceback (most recent call last):
  File "Training/tf/create_dataset.py", line 10, in <module>
    from utils.gen_preprocessing import compute_genmatch_dR, recompute_tau_type, dict_to_numba
  File "/afs/cern.ch/work/k/kandroso/tau/TauMLTools/Training/tf/utils/gen_preprocessing.py", line 3, in <module>
    import vector
ModuleNotFoundError: No module named 'vector'
```

shedprog (changes requested):

Training/tf/create_dataset.py
```diff
@@ -0,0 +1,126 @@
...
+                data = ()
+
+                # add awkward arrays converted to TF ragged arrays
+                for particle_type, feature_names in cfg['feature_names'].items(): # do this separately for each particle collection
+                    X = awkward_to_tf(a_selected, particle_type, feature_names) # will keep only feats from feature_names
+                    data += (X,)
+                
+                # all labels to final dataset
+                y = ak.to_pandas(a_selected[cfg["label_columns"]]).values
+                data += (y,)
```
shedprog, line 93: I am thinking that storing together tf.array and np.array in one tuple might not be a good way. Also, we might need to check if preliminary converting of X to tensorflow gives better performance than just directly `from_slices((numpy, numpy,))`

Training/tf/create_dataset.py
```diff
@@ -0,0 +1,126 @@
...
+
+                # remove existing datasets
+                path_to_dataset = to_absolute_path(f'{cfg["path_to_dataset"]}/{cfg["dataset_name"]}/{dataset_type}/{os.path.splitext(os.path.basename(file_name))[0]}/{tau_type}')
+                if os.path.exists(path_to_dataset):
+                    shutil.rmtree(path_to_dataset)
+                else:
+                    os.makedirs(path_to_dataset, exist_ok=True)
+
+                # save
+                tf.data.experimental.save(dataset, path_to_dataset)
```
shedprog, line 114: According to the new release docs, they now moved it from experimental to the tf.data class methods: https://www.tensorflow.org/api_docs/python/tf/data/Dataset#save

shedprog (changes requested):
Comments on the **ROOT -> tf.data preprocessing** part of the new data loading method

PR #135, opened 2022-12-22
Fixed obsolete EDAnalyzer warnings + workaround for TauPOGSpinner crashes

kandrosov:
- Switched to new EDAnalyzers. using `one` variant because current ntuple filling code is not parallelizable.
- TauPOGSpinner currently crashes for ttbar. Added option to disable it. Keep spinner weights only for Higgs and DY samples. In the future, we should consider refactoring of TauPOGSpinner code to improve stability and readability.

### DIFF
--- a/Production/plugins/TauTupleProducer.cc
+++ b/Production/plugins/TauTupleProducer.cc
@@ -245,9 +245,9 @@ private:
         tauTuple().deepmet_phi = deepMETs->at(0).phi();
         tauTuple().genmet_pt = genMETs->at(0).pt();
         tauTuple().genmet_phi = genMETs->at(0).phi();
-        tauTuple().tauSpinnerWTEven = tauSpinnerWTEven.isValid() ? *tauSpinnerWTEven : -1.f;
-        tauTuple().tauSpinnerWTOdd = tauSpinnerWTOdd.isValid() ? *tauSpinnerWTOdd : -1.f;
-        tauTuple().tauSpinnerWTMM = tauSpinnerWTMM.isValid() ? *tauSpinnerWTMM : -1.f;
+        tauTuple().tauSpinnerWTEven = tauSpinnerWTEven.isValid() ? *tauSpinnerWTEven : default_value;
+        tauTuple().tauSpinnerWTOdd = tauSpinnerWTOdd.isValid() ? *tauSpinnerWTOdd : default_value;
+        tauTuple().tauSpinnerWTMM = tauSpinnerWTMM.isValid() ? *tauSpinnerWTMM : default_value;
 
         TauJetBuilder builder(builderSetup, *taus, *boostedTaus, *jets, *fatJets, *cands, *electrons, *photons, *muons,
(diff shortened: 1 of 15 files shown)


Production/plugins/TauTupleProducer.cc
```diff
@@ -66,86 +66,33 @@ struct TauTupleProducerData {
...
-        n_producers(0)
-    {
-        summaryTuple().numberOfProcessedEvents = 0;
-    }
-
-    ~TauTupleProducerData() {}
-
-public:
-
-    static TauTupleProducerData* RequestGlobalData()
```
ofivite, line 89: Just for my education, why the methods `*GlobalData` were used before? (and not needed now)

Production/plugins/TauTupleProducer.cc
```diff
@@ -289,28 +233,28 @@ class TauTupleProducer : public edm::EDAnalyzer {
...
+        tauTuple().deepmet_pt = deepMETs->at(0).pt();
+        tauTuple().deepmet_phi = deepMETs->at(0).phi();
         tauTuple().genmet_pt = genMETs->at(0).pt();
         tauTuple().genmet_phi = genMETs->at(0).phi();
-	tauTuple().tauSpinnerWTEven = (*tauSpinnerWTEven);
-        tauTuple().tauSpinnerWTOdd = (*tauSpinnerWTOdd);
-        tauTuple().tauSpinnerWTMM = (*tauSpinnerWTMM);
+        tauTuple().tauSpinnerWTEven = tauSpinnerWTEven.isValid() ? *tauSpinnerWTEven : -1.f;
+        tauTuple().tauSpinnerWTOdd = tauSpinnerWTOdd.isValid() ? *tauSpinnerWTOdd : -1.f;
+        tauTuple().tauSpinnerWTMM = tauSpinnerWTMM.isValid() ? *tauSpinnerWTMM : -1.f;
```
ofivite, line 250: Do I understand correctly that the workaround for crashes is this `isValid` check? And for ttbar mostly it isn't valid?

kandrosov: `isValid` is method of Handle. If handle is not valid it means that the corresponding producer is not run. Which is the case if `runTauSpinner=False`.

Production/plugins/TauTupleProducer.cc
kandrosov: It comes from a custom legacy code. It was an attempt to improve performance by parallelising producer, but we end up having `lock(mutex)` in the beginning of `produce` method, so it never really worked. If in the future we decide to implement multi-threading support during tuple production, we can use standard CMSSW interfaces for this.

shedprog:
I would merge this PR asap since without it the Production soft can not be compiled.

kandrosov:
> I would merge this PR asap since without it the Production soft can not be compiled.

with the last commit, the code in this PR should compile. PR is ready to be merged, as far as I see.

ofivite (approved):

PR #137, opened 2022-12-30
Added possibility to use regexp in Smart Trees

shedprog:
Hi all,

in this PR I am trying to add some functionality for disabling branches on the production step (needed to preserve the space for displaced tau study when only pfCands are needed), but also might be useful for other studies where only a small part of the variable list in TauTuple.h is required. I am trying to add modifications in two steps:

1) Possibility to use regular expression for enabling and disabling the branches. Now, in addition to exact matching e.g: “tau_pt” regular expressions are allowed, like: “boostedTau_.*”, matching is performed with std::regex_match, so it attempts to match a regular expression to an entire character sequence (branch name) what should prevent the situations where e.g: “tau_pt” would match “tau_pt_weighted_deta_strip” and e.t.c

2) Just disabling tree linkage but leaving declaration of these variables for the writing mode does not work. Because, in the `get` method that is extensively used in tuple production it uses SmartTreeEntryMap of string to the tau tuple variable [object](https://github.com/cms-tau-pog/TauMLTools/blob/7368a01775f894cc9b3e1770e49417c2ab2dac9d/Core/interface/SmartTree.h#L390-L395) in this way `get` method is more safe than `tauTuple().branch_name` since it throw an error on excess request when variable is not linked. However, it does not allow me to use the same TauTupleProducer.cc. The easiest solution would be to add also disabled branches to [`SmartTreeEntryMap entries`](https://github.com/cms-tau-pog/TauMLTools/blob/9342a995dcd05dd1ab62483d6580a804ad280507/Core/interface/SmartTree.h#L336) since it is not  really used anywhere except `get`, however it reduces some safety. Would such compromise be acceptable?

### DIFF
--- a/Core/interface/SmartTree.h
+++ b/Core/interface/SmartTree.h
@@ -121,7 +121,7 @@ namespace detail {
     template<typename DataType>
     struct BranchCreator {
         void Create(TTree& tree, const std::string& branch_name, DataType& value, bool readMode,
-                    SmartTreeEntryMap& entries, bool disabledBranch)
+                    SmartTreeEntryMap& entries, bool enabled)
         {
             using FixesMap = std::unordered_map<std::string, std::string>;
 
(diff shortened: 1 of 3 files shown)


Core/interface/SmartTree.h
```diff
@@ -120,7 +121,7 @@ namespace detail {
     template<typename DataType>
     struct BranchCreator {
         void Create(TTree& tree, const std::string& branch_name, DataType& value, bool readMode,
-                    SmartTreeEntryMap& entries)
+                    SmartTreeEntryMap& entries, bool disabledBranch)
```
kandrosov, line 124:
```suggestion
                    SmartTreeEntryMap& entries, bool enabled)
```

Core/interface/SmartTree.h
```diff
@@ -147,40 +148,45 @@ namespace detail {
                 throw std::runtime_error("Entry is already defined.");
             entries[branch_name] = entry;
 
-            TClass *cl = TClass::GetClass(typeid(DataType));
-
-            if(readMode) {
-                try {
-                    EnableBranch(tree, branch_name);
-
+            if(!disabledBranch) {
```
kandrosov, line 151:
```suggestion
if(!enabled) {
    std::cerr << "Warning: SmartTree branch '" << branch_name << "' will be disabled." << std::endl;
    return;
}
```

Core/interface/SmartTree.h
```diff
@@ -312,11 +318,20 @@ class SmartTree {
     void AddBranch(const std::string& branch_name, DataType& value)
     {
         std::lock_guard<Mutex> lock(mutex);
-        if (!disabled_branches.count(branch_name) && (!enabled_branches.size() || enabled_branches.count(branch_name))){
-            detail::BranchCreator<DataType> creator;
-            creator.Create(*tree, branch_name, value, readMode, entries);
+        auto exp_match = [](std::set<std::string> matching_branches, std::string branch_name) -> bool {
+            for (std::string matching_branch: matching_branches){
```
kandrosov, line 322:
```suggestion
        auto exp_match = [](const std::set<std::string>& matching_branches, const std::string& branch_name) -> bool {
            for (const std::string& matching_branch : matching_branches){
```

Core/interface/SmartTree.h
```diff
@@ -312,11 +318,20 @@ class SmartTree {
...
+            return false;
+        };
+        detail::BranchCreator<DataType> creator;
+        if ((!disabled_branches.size() || !exp_match(disabled_branches,branch_name)) &&
+            (!enabled_branches.size()  || exp_match(enabled_branches,branch_name))) {
+            creator.Create(*tree, branch_name, value, readMode, entries, false);
             active_branches.insert(branch_name);
-        }
+        } else
+            creator.Create(*tree, branch_name, value, readMode, entries, true);
```
kandrosov, line 334:
```c++
const bool enabled = (!disabled_branches.size() || !exp_match(disabled_branches,branch_name)) &&
    (!enabled_branches.size()  || exp_match(enabled_branches,branch_name));
if(enabled)
    active_branches.insert(branch_name);
creator.Create(*tree, branch_name, value, readMode, entries, enabled);
```

Production/plugins/TauTupleProducer.cc
```diff
@@ -69,13 +69,16 @@ struct TauTupleProducerData {
     std::unique_ptr<tau_tuple::TauTuple> tauTuple;
     std::unique_ptr<tau_tuple::SummaryTuple> summaryTuple;
 
-    TauTupleProducerData() :
+    TauTupleProducerData(const std::vector<std::string>& disableBranches) :
```
kandrosov, line 72: for completeness, add also enabledBranches

kandrosov (changes requested):
The current solution looks fine for me. In the future we can consider adding `get`-like method that creates branch at the first call and use only it at the production stage (e.g. no more `tauTuple().variable`. In this way we can remove macro-based definition of TauTuple in `TauTuple.h`, which is no longer relevant because we generate TauTuple from TTree at the training step.

kandrosov (approved):

PR #138, opened 2023-01-03
Fix ReReco at MiniAOD, added TauJetTag selector (Displaced) 

shedprog:
Hi All,

In this PR: little customization for displaced tau reconstruction is added and few inconsistencies for tauReco at MiniAOD level is fixed. For the description of PR parts, please read below:

1) For the compatibility with current tupling working flow—selector is added with the naming `TauJetTag`. `TauJetTag` selects the events with the condition of having genLepton/genJet matched to the Jet/fatJet. Some space is left to add gen object selection criteria in the future PRs. This selector is needed to (a) to drop events with only boostedTau (since it is not supposed to be filled in some cases not to have empty events) (b) to have better control over matching then just using requireGenMatch flag. Also, it might be interesting to further have a Jet-Based selector in case we switch to Jet tagging at some point, so might be relevant not only for displaced. This requires discussion.

2) There was some modification of [`adaptToRunAtMiniAOD module`](https://github.com/cms-sw/cmssw/blob/master/RecoTauTag/Configuration/python/tools/adaptToRunAtMiniAOD.py) so I modified correspondingly the way of calling this and added it to the path (why before we did not need it? Because it was substituting the default path name?)

3) Setups for the `loose HPS` reconstruction is added in separate file containing Ricardo’s setups + what Sara did for HLT.

### DIFF
--- a/Production/python/Production.py
+++ b/Production/python/Production.py
@@ -4,7 +4,6 @@ import re
 import importlib
 import FWCore.ParameterSet.Config as cms
 from FWCore.ParameterSet.VarParsing import VarParsing
-import RecoTauTag.Configuration.tools.adaptToRunAtMiniAOD as tauAtMiniTools
 from RecoMET.METPUSubtraction.deepMETProducer_cfi import deepMETProducer
 import os
 
@@ -47,8 +46,10 @@ options.register('applyRecoPtSieve', False, VarParsing.multiplicity.singleton, V
                  "Randomly drop jet->tau fakes depending on reco tau pt to balance contributions from low and higt pt.")
 options.register('reclusterJets', True, VarParsing.multiplicity.singleton, VarParsing.varType.bool,
                 " If 'reclusterJets' set true a new collection of uncorrected ak4PFJets is built to seed taus (as at RECO), otherwise standard slimmedJets are used")
-options.register('rerunTauReco', False, VarParsing.multiplicity.singleton, VarParsing.varType.bool,
-                "If true, tau reconstruction is re-run on MINIAOD with a larger signal cone and no DM finding filter")
+options.register('rerunTauReco', '', VarParsing.multiplicity.singleton, VarParsing.varType.string,
+                """If not empty, tau reconstruction is re-run on MINIAOD (available modes:
+                   1) signalCone - larger signal cone and no DM finding filter,
+                   2) displacedTau - loose max Impact Parameter and signal quality cuts)""")
 options.register('runTauSpinner', False, VarParsing.multiplicity.singleton, VarParsing.varType.bool,
                  "Run TauPOGSpinner to store CP weights")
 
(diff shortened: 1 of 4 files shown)


Production/python/Production.py
```diff
@@ -133,31 +134,38 @@
 if options.rerunTauReco:
     tau_collection = 'selectedPatTaus'
 
-    tauAtMiniTools.addTauReReco(process)
-    tauAtMiniTools.adaptTauToMiniAODReReco(process, options.reclusterJets)
+    tauAtMiniConfig = importlib.import_module('RecoTauTag.Configuration.tools.adaptToRunAtMiniAOD')
```
kandrosov, line 137: why `importlib.import_module`? simple import doesn't work for some reason?

Production/python/setupTauReReco.py
```diff
@@ -0,0 +1,50 @@
...
+    process.selectedPatTaus.cut = cms.string('pt > 18.')   ## remove DMFinding filter (was pt > 18. && tauID(\'decayModeFindingNewDMs\')> 0.5)
+
+def reReco_DisTau(process):
+
+    # The following study is considered as a starting point for modifications:
+    # https://indico.cern.ch/event/868576/contributions/3696661/attachments/1979084/3295013/tau_displaced_ll_ws_31_1_2020.pdf
+
+    process.selectedPatTaus.cut = cms.string('pt > 18.')   ## remove DMFinding filter (was pt > 18. && tauID(\'decayModeFindingNewDMs\')> 0.5)
+
+    process.combinatoricRecoTaus.builders[0].qualityCuts.signalQualityCuts.maxDeltaZ = cms.double(100.)
```
kandrosov, line 23: Are you applied the same changes to the cuts for all builders? if yes, I suggest using `for builder in process.combinatoricRecoTaus.builder ...`

Production/src/Selectors.cc
```diff
@@ -160,6 +162,24 @@ TauJetSelector::Result genTauTau::Select(const edm::Event& event, const std::deq
...
+                                     const reco::Vertex& primaryVertex,
+                                     const pat::TriggerObjectStandAloneCollection& triggerObjects,
+                                     const edm::TriggerResults& triggerResults, float rho)
+{
+    std::vector<const TauJet*> selected;
+    for(const TauJet& tauJet :tauJets) {
+      const ObjPtr<reco_tau::gen_truth::GenLepton>& genLepton = tauJet.genLepton;
+      const ObjPtr<const reco::GenJet>& genJet = tauJet.genJet;
+      const ObjPtr<const pat::Jet>& jet = tauJet.jet;
+      if(genLepton || genJet || jet)
```
kandrosov, line 177:
```suggestion
      if(tauJet.genLepton || tauJet.genJet || tauJet.jet)
```

Production/src/Selectors.cc
```diff
@@ -160,6 +162,24 @@ TauJetSelector::Result genTauTau::Select(const edm::Event& event, const std::deq
...
+        selected.push_back(&tauJet);
+    }
+    if(selected.empty()) return {};
```
kandrosov, line 180:
```suggestion
```
or something breaks without this line?..

kandrosov (changes requested):

Production/python/Production.py
shedprog: It also works as `import RecoTauTag.Configuration.tools.adaptToRunAtMiniAOD as tauAtMiniConfig`, would it be more propriety way? I have seen that ID module is loaded this way and followed the same way.

kandrosov: It is a legacy code. At some point we was using modified tauID customisation module in nonstandard location, so `import` didn't work. Currently, I think we should be able to get rid of `importlib` everywhere in `Production.py`.

kandrosov (approved):

PR #140, opened 2023-02-10
TF preprocessing

ofivite:
This PR is the first split of https://github.com/cms-tau-pog/TauMLTools/pull/133, describing the workflow to process ROOT files into `tf.data`. Conceptually, comparing with https://github.com/cms-tau-pog/TauMLTools/pull/133 it is updated with the switch to `uproot.dask`[[1](https://uproot.readthedocs.io/en/latest/uproot._dask.dask.html), [2](https://github.com/kkothari2001/PyHEP2022-uproot.dask)] and comments from the previous review.

Specifically: 

- Lazy computations with `dask`. This gives better memory control on the awkward side (with `step_size` parameter) which can be performed lazily in chunks. Also, no input branches is needed now since they are read on demand [[comment](https://github.com/cms-tau-pog/TauMLTools/pull/133#discussion_r1038025597)]. Unfortunatelly, `ak.where()` doesn't support lazy computations, so features with this transformation have to be computed beforehand. Otherwise, graphs are constructed on the per feature level and computed within `awkward_to_tf` function. 
- Removing of `tau_type` split. The split of the output into separate TF datasets per class is removed and therefore no class selection is done at this step. This will happen at the training and evaluation steps.  
- Support of `xrootd` paths. Tim commented that `to_absolute_path()` is not compatible with xrootd, so separate handling (including globing of remote directories) is added. NB: filename pattern are not supported, path to remote directory should be specified instead and it will read _all_ the ROOT files from there.
- Adding compression of the output TF datasets.
- Update of conda environment. This corresponds to:

```
conda upgrade  seaborn numpy numba pandas
pip install --upgrade uproot awkward dask-awkward vector tensorflow hydra-core mlflow scipy
pip install jupyterlab
```
- Addressed comments [[1](https://github.com/cms-tau-pog/TauMLTools/pull/133#discussion_r1038036323), [2](https://github.com/cms-tau-pog/TauMLTools/pull/133#discussion_r1038171116), [3](https://github.com/cms-tau-pog/TauMLTools/pull/133#discussion_r1038316552), [4](https://github.com/cms-tau-pog/TauMLTools/pull/133#discussion_r1038365070)]
- Minor refactoring.

CPU memory profiling with 1 S&M file looks as follows. On my benchmark, the peak value is 10-20% less comparing to the previous implementation.
image


tau-ml-env.yaml
```diff
@@ -361,36 +348,90 @@ dependencies:
...
+      - dask-awkward==2023.1.0
+      - flatbuffers==2.0
+      - fqdn==1.5.1
+      - fsspec==2023.1.0
+      - gast==0.4.0
+      - google-auth==2.6.6
+      - google-auth-oauthlib==0.4.6
+      - google-pasta==0.2.0
+      - grpcio==1.46.1
+      - h5py==3.6.0
```
kandrosov, line 372: please, prefer condo-based packages. Use pip only for packages that are not available or outdated in conda channels.

kandrosov (changes requested):
- I propose moving data preparation related code into a separate folder, while `Training` will contain only the actual training-related files.

ofivite:
> * I propose moving data preparation related code into a separate folder, while `Training` will contain only the actual training-related files.

Do you mean creating a new directory in the root (e.g. `TauMLTools/Preprocessing`)?

kandrosov:
> Do you mean creating a new directory in the root (e.g. `TauMLTools/Preprocessing`)?

Yes. And I think in the future (not in this PR), some code from `Analysis` folder should also migrate to `Preprocessing`.

tau-ml-env.yaml
ofivite: I've tried now to move some packages to `conda update`, starting from the current `tau-ml` environment. It looks like it either doesn't update it,  or the version is not the latest and greatest. Is it a problem to keep some of the packages `pip`-ed?  

I pushed 5f21004 where I did a pip-minimalistic update of the environment (starting from the master branch), so only mlflow and uproot+awkward are moved from conda to pip:
```
conda update conda
conda install -c conda-forge jupyterlab vector
conda update seaborn numpy numba pandas scipy
pip install --upgrade uproot awkward dask-awkward tensorflow hydra-core mlflow 
```

kandrosov: conda is better with managing consistency of the setup - with part of packages installed from conda and another part from pip, it is easy to end-up with a situation where something is broken. For that reason, I usually try to minimise the amount of packages installed with pip. But if you tested and the resulting environment works fine both on gpu-cluster and lxplus, I'm fine with this.

kandrosov (approved):

ofivite:
Added in fe97dad3c8770cb07b9c79039ef9abbde9ddaa11 also a check for `inf` values (found now in `muon_normalizedChi2` during the training tests)

ofivite:
a804e74dd33be4e5a8fb3a95e51824f758f900ca: downgraded TF to v2.10.1 in the conda environment since the previous setup was causing issues with Tensorboard.

PR #141, opened 2023-02-14
Generalization of the pre-processing (Displaced)

shedprog:
In the following PR modification for the displaced tau tagger workflow are added:

- [x] Added option to run crab on limited amount of units.
- [x] Added DisTauTagSelection.h with the displaced tau classes definition
- [x] Modified Analysis/bin/CreateSpectralHists.cxx to be able to run with boostedTau, tau, and Jet spectrum
- [x] Added simple script to merge all the files without spectrum control as an alternative to S&M
- [x] Organize documentation folder for supplementary developments
- [x] Updated preprocessing workflow documentation for displaced taus


Analysis/bin/CreateSpectralHists.cxx
```diff
@@ -41,21 +44,167 @@ struct HistArgs {
...
+  using Tau = tau_tuple::Tau;
+  using TauTuple = tau_tuple::TauTuple;
+
+  HistSpectrum(std::shared_ptr<TFile>& outputFile, const HistArgs& hargs) : root_ext::AnalyzerData(outputFile) {
+    eta_pt_hist.SetMasterHist(
+      hargs.eta_bins, hargs.eta_min, hargs.eta_max,
+      hargs.pt_bins, hargs.pt_min, hargs.pt_max);
+  };
+
+  virtual ~HistSpectrum() {};
```
kandrosov, line 60:
```suggestion
```

Analysis/bin/CreateSpectralHists.cxx
```diff
@@ -24,6 +25,8 @@ struct Arguments {
     run::Argument<std::string> exclude_list{"exclude-list", "comma separated list of files to exclude", ""};
     run::Argument<std::string> exclude_dir_list{"exclude-dir-list",
                                                "comma separated list of directories to exclude", ""};
+    run::Argument<std::string> mode{"mode",
```
kandrosov, line 28: can we run for all modes in one go?

Analysis/bin/SimpleMergeMix.cxx
```diff
@@ -0,0 +1,426 @@
...
+    run::Argument<unsigned> n_threads{"n-threads", "number of threads", 1};
+    run::Argument<unsigned> seed{"seed", "random seed to initialize the generator used for sampling", 1234567};
+    run::Argument<std::string> disabled_branches{"disabled-branches", "list of branches to disabled in the input tuples", ""};
+    run::Argument<std::string> compression_algo{"compression-algo","ZLIB, LZMA, LZ4","LZMA"};
+    run::Argument<unsigned> compression_level{"compression-level", "compression level of output file", 9};
+    run::Argument<unsigned> job_idx{"job-idx", "index of the job (starts from 0)"};
+    run::Argument<unsigned> n_jobs{"n-jobs", "the number by which to divide all files"};
+};
+
+struct SourceDesc {
```
kandrosov, line 83: there is a lot of code duplication with S&M code. Could the common part be extracted in .h and merging logic is customised with inheritance and virtual functions?

kandrosov (changes requested):

Analysis/bin/CreateSpectralHists.cxx
shedprog: About this I am a bit confused, why it gives me a warning like `has virtual functions and accessible non-virtual destructor`, in my understanding it is to prevent forgetting of calling the destructor for the `back to base class cast` like `Base* b = new Derived();` but you mean that in our case we are safe because we're only defining Derived class, right?

Analysis/bin/SimpleMergeMix.cxx
shedprog: No, at the moment I did not add this functionality, for instance DisplacedTau definition does not valid for any analysis but would just require more time to create spectrum hists and also initializing of some extra variables. I propose every analysis carefully checks its own definition.

Analysis/bin/CreateSpectralHists.cxx
shedprog: No, at the moment I did not add this functionality, for instance DisplacedTau definition does not valid for any analysis but would just require more time to create spectrum hists and also initializing of some extra variables. I propose every analysis carefully checks its own definition.

Analysis/bin/CreateSpectralHists.cxx
kandrosov: If the base class has a virtual destructor, there should be no need to define empty virtual destructor in child classes as it must be created by default by the compiler. And `AnalyzerData` [has one](https://github.com/cms-tau-pog/TauMLTools/blob/master/Core/interface/AnalyzerData.h#L85). I'm puzzled why would you are getting this warning... it is still there if you define `HistSpectrum` as `class`?

shedprog: My bad, I am saying in general, for instance for my SourceDesc<T> class, but now I understood, thank you.

kandrosov: Yes, in general, if class has virtual functions, you should also define a virtual destructor. But doing it once in the base class is sufficient.

kandrosov (dismissed):

PR #142, opened 2023-03-02
Switching to a new environment definition

kandrosov:
- added RunKit submodule that will be used for crab production 
- using cmssw-independent base environment (conda or lcg)
- to commands in CMSSW environment use `cmsEnv` (`cmsEnv11_2` or `cmsEnv11_3` for phase2)
- moving towards a better integration of TauML workflow with law
   - always load law in env.sh
   - moved Analysis/law into the root directory


README.md
```diff
@@ -333,31 +330,31 @@ then run the job submission
...
+where ```--version``` specifies the job submission version (used by law to monitor the status of the jobs), ```--environment``` specifies the shell environment to be used (*conda* in this case, see [TauMLTools/Analysis/law/bootstrap.sh](https://github.com/cms-tau-pog/TauMLTools/blob/master/Analysis/law/bootstrap.sh)), ```--cfg``` is the yaml configuration file (the same used in interactive mode), ```--files-per-job``` specifies the number of files processed by each job, ```--n-jobs``` specifies the number of jobs to run, and ```--output-path``` specifies the path to the output directory, used to save the results.
```
ofivite, line 333:
```suggestion
where ```--version``` specifies the job submission version (used by law to monitor the status of the jobs), ```--environment``` specifies the shell environment to be used (*conda* in this case, see [TauMLTools/bootstrap.sh](https://github.com/cms-tau-pog/TauMLTools/blob/master/bootstrap.sh)), ```--cfg``` is the yaml configuration file (the same used in interactive mode), ```--files-per-job``` specifies the number of files processed by each job, ```--n-jobs``` specifies the number of jobs to run, and ```--output-path``` specifies the path to the output directory, used to save the results.
```

ofivite: \+ please also edit setup instructions in a few lines above (L324 in the PR version of the file)

env.sh
```diff
@@ -21,55 +21,84 @@ run_cmd() {
...
+    #for phase2
+    run_cmd install_cmssw slc7_amd64_gcc900 CMSSW_11_2_5 7
+    run_cmd install_cmssw slc7_amd64_gcc900 CMSSW_11_3_0 7
+
+    alias cmsEnv="env -i HOME=$HOME ANALYSIS_PATH=$ANALYSIS_PATH ANALYSIS_DATA_PATH=$ANALYSIS_DATA_PATH X509_USER_PROXY=$X509_USER_PROXY DEFAULT_CMSSW_BASE=$DEFAULT_CMSSW_BASE $ANALYSIS_PATH/RunKit/cmsEnv.sh"
+    alias cmsEnv11_2="env -i HOME=$HOME ANALYSIS_PATH=$ANALYSIS_PATH ANALYSIS_DATA_PATH=$ANALYSIS_DATA_PATH X509_USER_PROXY=$X509_USER_PROXY DEFAULT_CMSSW_BASE=$ANALYSIS_PATH/soft/CentOS$os_version/CMSSW_11_2_5 $ANALYSIS_PATH/RunKit/cmsEnv.sh"
+    alias cmsEnv11_3="env -i HOME=$HOME ANALYSIS_PATH=$ANALYSIS_PATH ANALYSIS_DATA_PATH=$ANALYSIS_DATA_PATH X509_USER_PROXY=$X509_USER_PROXY DEFAULT_CMSSW_BASE=$ANALYSIS_PATH/soft/CentOS$os_version/CMSSW_11_3_0 $ANALYSIS_PATH/RunKit/cmsEnv.sh"
+  fi
+
+  if [[ $MODE == *"conda"* ]]; then
```
ofivite, line 101: Would it make sense to update to the latest version of `law` in conda? (currently we are with `0.1.5`, but the latest is [`0.1.12`](https://github.com/riga/law/releases/tag/v0.1.12)). But I am not sure about the advantages of this, as well as about the breaking changes.

env.sh
```diff
@@ -21,55 +21,84 @@ run_cmd() {
...
+
+  if [[ $MODE = *"cmssw"* ]]; then
+    run_cmd install_cmssw slc7_amd64_gcc10 $default_cmssw_ver 7
+    run_cmd install_cmssw el8_amd64_gcc10 $default_cmssw_ver 8
+
+    #for phase2
+    run_cmd install_cmssw slc7_amd64_gcc900 CMSSW_11_2_5 7
+    run_cmd install_cmssw slc7_amd64_gcc900 CMSSW_11_3_0 7
+
+    alias cmsEnv="env -i HOME=$HOME ANALYSIS_PATH=$ANALYSIS_PATH ANALYSIS_DATA_PATH=$ANALYSIS_DATA_PATH X509_USER_PROXY=$X509_USER_PROXY DEFAULT_CMSSW_BASE=$DEFAULT_CMSSW_BASE $ANALYSIS_PATH/RunKit/cmsEnv.sh"
```
ofivite, line 96: For my understanding, what would be the idea behind these aliases? So that they automatically source the relevant CMSSW environment prior to the execution of some command?

ofivite (changes requested):

env.sh
kandrosov: yes, so the base environment do not have any CMSSW contamination. For that reason we need some "base environment". It could be conda or lcg. I put lcg by default, because it is way faster on lxplus. Then `conda_cmssw` would mean "use conda as the base environment & enable possibility to run cmssw". While for `source env.sh conda` or `source env.sh` cmssw releases are not installed (which is useful for trainings on non cms clusters)

ofivite: OK, makes sense!

ofivite (dismissed):

PR #143, opened 2023-03-13
Fix bug in running customiseHLT on data samples + run splitting data/MC 

zhiyuanlcern:


Production/plugins/TauTableProducerHLT.cc
```diff
@@ -29,65 +29,77 @@ class TauTableProducerHLT : public edm::global::EDProducer<> {
...
+        if (hasSecondaryVertex[tau_index] > 0) {
+          secondaryVertex_x[tau_index] = tausIPHandle->value(tau_index)->secondaryVertex()->x();
+          secondaryVertex_y[tau_index] = tausIPHandle->value(tau_index)->secondaryVertex()->y();
+          secondaryVertex_z[tau_index] = tausIPHandle->value(tau_index)->secondaryVertex()->z();
+        }
+        else {
+          secondaryVertex_x[tau_index] = -999.;
+          secondaryVertex_y[tau_index] = -999.;
+          secondaryVertex_z[tau_index] = -999.;
+        };
```
kandrosov, line 96: it will be set to `default_value` at the initialisation.
```suggestion
```

kandrosov (changes requested):

Production/python/customiseHLT.py
```diff
@@ -315,15 +317,32 @@ def customise(process):
...
+                                     process.tauTablesTask,
+                                     process.tauTablesExtTask,
+                                     process.pfCandTablesTask,
+                                     process.AK4PFJetsTableTask,
+                                     process.recoAllGenJetsNoNuTask,
+                                     process.L1TableTask,
+                                     process.caloTableTask,
+                                     process.pixelTrackTableTask)
+
+  is_data =  True
```
ofivite, line 330: Should this assignment be removed? Otherwise the function will always run in `is_data = True` mode.

ofivite (changes requested):

kandrosov (approved):

PR #144, opened 2023-03-15
Review of HLT nano production code

kandrosov:
- prepared data and MC HLT configuration in a crab-compatible form
- defined crab submission configs for HLT
- variable_check.py: updated to new uproot interface
- Update submission and skimming tools in RunKit
- cms-merge-topic for customisation function for 2023 PFCluster thresholds at HLT
- HLT nano:
  - added primary and secondary PF vertices
  - pruned constant and redundant variables
  - store extended information for L1 taus
  - added skim step on cmssw output: keep only rechits and pf candidation that are within dR to L1 tau, Jet or HPS Tau with pt > 15 GeV and |eta| < 2.7
- updated documentation


Production/crab/Run3Winter23_HLT/data.yaml
```diff
@@ -0,0 +1,15 @@
...
+EphemeralHLTPhysics1_Run2022G: /EphemeralHLTPhysics1/Run2022G-v1/RAW
+EphemeralHLTPhysics2_Run2022G: /EphemeralHLTPhysics2/Run2022G-v1/RAW
+EphemeralHLTPhysics3_Run2022G: /EphemeralHLTPhysics3/Run2022G-v1/RAW
+EphemeralHLTPhysics4_Run2022G: /EphemeralHLTPhysics4/Run2022G-v1/RAW
+EphemeralHLTPhysics5_Run2022G: /EphemeralHLTPhysics5/Run2022G-v1/RAW
+EphemeralHLTPhysics6_Run2022G: /EphemeralHLTPhysics6/Run2022G-v1/RAW
+EphemeralHLTPhysics7_Run2022G: /EphemeralHLTPhysics7/Run2022G-v1/RAW
+EphemeralHLTPhysics8_Run2022G: /EphemeralHLTPhysics8/Run2022G-v1/RAW
+ZeroBias_Run2022G: /ZeroBias/Run2022G-v1/RAW
+
```
shedprog, line 15: Just general comment: Would it make sense to separate at the moment all the files with old configuration (*.txt) vs. *.yaml in different folders, for instance `crab` and `crab_runkit`

Analysis/python/variable_check.py
```diff
@@ -6,34 +6,30 @@
 parser.add_argument('--tree', required=True, type=str, help="Tree name")
 parser.add_argument('--max-entries', required=False, type=int, default=None,
                     help="Maximal number of entries to process")
+parser.add_argument('--std-thr', required=False, type=float, default=1e-7,
+                    help="If std(var) < std_thr, var is considered constant.")
```
shedprog, line 10: Asking out of my curiosity: do I understand correctly that then you will not consider these variables for the training at all?

Production/crab/Run3Winter23_HLT/data.yaml
kandrosov: are we going to use old files in any of the next productions? I think after the new tau-tuple production for 2018 is done, we can clean-up old txt files (they can be still accessible in git history)

Analysis/python/variable_check.py
kandrosov: yes, this script was used to prune const variables before the production, so they are not included in the PR and will not be used for the training.

PR #145, opened 2023-03-23
New selection process

msessini:
First proposal to open discussion about the implementation of a new selector as suggested by Konstantin :

https://indico.cern.ch/event/1263238/contributions/5305055/attachments/2606494/4502197/2023-03-07%20TauTuple%20selection.pdf

### DIFF
--- a/Analysis/interface/TauTuple.h
+++ b/Analysis/interface/TauTuple.h
@@ -137,6 +137,7 @@
     VAR(Int_t, genJet_partonFlavour) /* parton-based flavour */ \
     /* Tag object variables (for tag-and-probe data) */ \
     VAR(Int_t, tagObj_valid) /* indicates presence of the tag object */ \
+    VAR(Int_t, tagObj_kind) /* 1 Muon 2 Electron */ \
     VAR4(Float_t, tagObj_pt, tagObj_eta, tagObj_phi, tagObj_mass) /* 4-momentum of the tag object */ \
     VAR(Int_t, tagObj_charge) /* charge of the tag object */ \
     VAR(UInt_t, tagObj_id) /* ID of the tag object */ \
--- a/Production/plugins/TauTupleProducer.cc
+++ b/Production/plugins/TauTupleProducer.cc
@@ -261,6 +261,7 @@ private:
         const auto [tauJets, tagObj] = selector->Select(event, builder.GetTauJets(), *electrons, *muons,
                                                            METs->at(0), PV, *triggerObjects, *triggerResults, *rho);
         tauTuple().tagObj_valid = tagObj != nullptr;
+        tauTuple().tagObj_kind = tagObj ? static_cast<int>(tagObj->kind) : default_int_value;
         tauTuple().tagObj_pt = tagObj ? tagObj->p4.pt() : default_value;
         tauTuple().tagObj_eta = tagObj ? tagObj->p4.eta() : default_value;
         tauTuple().tagObj_phi = tagObj ? tagObj->p4.phi() : default_value;
--- a/Production/src/Selectors.cc
+++ b/Production/src/Selectors.cc
@@ -55,6 +55,8 @@ std::shared_ptr<TauJetSelector> TauJetSelector::Make(const std::string& name)
         return std::make_shared<genTauTau>();
     if(name == "TauJetTag")
         return std::make_shared<TauJetTag>();
+    if(name == "TagAndProbe") 
+        return std::make_shared<TagAndProbe>();
     throw analysis::exception("Unknown selector name = '%1%'") % name;
 }
 
@@ -273,7 +275,7 @@ TauJetSelector::Result TagAndProbe::Select(const edm::Event& event, const std::d
     //
     else { // No mu or ele tag nor high MET 
         for(const TauJet& tauJet : tauJets) {
-            if(!((tauJet.genLepton && tauJet.genJet) && !(std::abs(tauJet.genLepton->visibleP4().eta()) < 2.8 || std::abs(tauJet.genJet->eta()) < 2.8)) continue;
+            if(!(tauJet.genLepton && tauJet.genJet) && !(std::abs(tauJet.genLepton->visibleP4().eta()) < 2.8 || std::abs(tauJet.genJet->eta()) < 2.8)) continue;
 	    selectedTauJets.push_back(&tauJet); //Fill with jets passing selection
 	}
 	if(selectedTauJets.size() <= 2)
--- a/Production/src/Selectors.cc
+++ b/Production/src/Selectors.cc
@@ -177,5 +177,123 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
     return Result(selected, nullptr);
 }
 
+TauJetSelector::Result TagAndProbe::Select(const edm::Event& event, const std::deque<TauJet>& tauJets,
+                                     const std::vector<pat::Electron>& electrons,
+                                     const std::vector<pat::Muon>& muons, const pat::MET& met,
+                                     const reco::Vertex& primaryVertex,
+                                     const pat::TriggerObjectStandAloneCollection& triggerObjects,
+                                     const edm::TriggerResults& triggerResults, float rho)
+{
+    const pat::Muon *ref_muon = nullptr;
+    const pat::Electron *ref_electron = nullptr;
+    for(const pat::Muon& muon : muons) { // Look for a muon tag first
+        if(!(muon.pt() > 26 && std::abs(muon.eta()) < 2.1 && muon.isMediumMuon() && PFRelIsolation(muon) < 0.1
+                && std::abs(muon.muonBestTrack()->dz(primaryVertex.position())) < 0.2
+                && std::abs(muon.muonBestTrack()->dxy(primaryVertex.position())) < 0.045))
+            continue;
+        if(!ref_muon || PFRelIsolation(*ref_muon) > PFRelIsolation(muon) || (PFRelIsolation(*ref_muon) == PFRelIsolation(muon) && ref_muon->pt() < muon.pt())){
+            ref_muon = &muon;
+        }
+    }
+    if(!ref_muon) {
+    	for(const pat::Electron& electron : electrons) { // Start looking for an ele tag if no muon was found
+            if(!(electron.pt() > 35 && std::abs(electron.eta()) < 2.3 && electron.electronID("mvaEleID-Fall17-noIso-V2-wp90") > 0.5f
+                     && std::abs(electron.bestTrack()->dz(primaryVertex.position())) < 0.2
+                     && std::abs(electron.bestTrack()->dxy(primaryVertex.position())) < 0.045))
+            	 continue;
+            if(!ref_electron || ref_electron->pt() < electron.pt()){
+                ref_electron = &electron;
+            }
+        }
+    }
+    //
+    auto tagObject = std::make_shared<TagObject>();
+    std::vector<const TauJet*> selectedTauJets;
+    if(ref_muon) { // Fill tag object with muon if muon found
+        tagObject->kind = tau_analysis::selectors::TagObject::Kind::Muon; //new tag object member 
+        tagObject->p4 = ref_muon->polarP4();
+        tagObject->charge = ref_muon->charge();
+        tagObject->id = unsigned(ref_muon->isLooseMuon()) * 1 + unsigned(ref_muon->isMediumMuon()) * 2
+                          + unsigned(ref_muon->isTightMuon(primaryVertex)) * 4;
+        tagObject->isolation = PFRelIsolation(*ref_muon);
+        tagObject->has_extramuon = hasExtraMuon(muons, ref_muon, primaryVertex);
+        tagObject->has_extraelectron = hasExtraElectron(electrons, rho);
+        tagObject->has_dimuon = hasExtraDimuon(muons, primaryVertex);
+        const TauJet* selected = nullptr;
+        for(const TauJet& tauJet : tauJets) { //Look for highest pT jet
+            if(!tauJet.jet) continue;
+	    const pat::Jet& jet = *tauJet.jet;
+            if(!(std::abs(jet.eta()) < 2.8 && reco::deltaR(jet.polarP4(), ref_muon->polarP4()) > 0.8))
+		    continue;
+	    if(!selected || selected->jet->pt() < jet.pt()) {
+		    selected = &tauJet;
+		    selectedTauJets = { selected };
+            }
+	}
+	return Result(selectedTauJets, tagObject);
+    }
+    //	
+    else if(ref_electron) { // Fill tag with ele if ele found rather than muon
+	tagObject->kind = tau_analysis::selectors::TagObject::Kind::Electron; //new tag object member
+        tagObject->p4 = ref_electron->polarP4();
+        tagObject->charge = ref_electron->charge();
+        tagObject->id = ref_electron->electronID("mvaEleID-Fall17-noIso-V2-wp90"); //Not sure what to put here 
+        tagObject->isolation = PFRelIsolation(*ref_electron,rho);
+        tagObject->has_extramuon = hasExtraMuon(muons, ref_muon, primaryVertex);
+        tagObject->has_extraelectron = hasExtraElectron(electrons, rho);
+        tagObject->has_dimuon = hasExtraDimuon(muons, primaryVertex);	
+	const TauJet* selected = nullptr;
+        for(const TauJet& tauJet : tauJets) { //Look for highest pT jet
+            if(!tauJet.jet) continue;
+            const pat::Jet& jet = *tauJet.jet;
+            if(!(std::abs(jet.eta()) < 2.7 && reco::deltaR(jet.polarP4(), ref_electron->polarP4()) > 0.8))
+                    continue;
+            if(!selected || selected->jet->pt() < jet.pt()) {
+                    selected = &tauJet;
+		    selectedTauJets = { selected };
+            }
+        }
+        return Result(selectedTauJets, tagObject);
+    }
+    //
+    else if(met.pt() > 180) { // If no mu or ele tag, then look for event with MET>180GeV as third option
+	const TauJet* selected = nullptr;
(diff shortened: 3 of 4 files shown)


Production/src/Selectors.cc
```diff
@@ -177,5 +179,123 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
...
+                                     const std::vector<pat::Electron>& electrons,
+                                     const std::vector<pat::Muon>& muons, const pat::MET& met,
+                                     const reco::Vertex& primaryVertex,
+                                     const pat::TriggerObjectStandAloneCollection& triggerObjects,
+                                     const edm::TriggerResults& triggerResults, float rho)
+{
+    const pat::Muon *ref_muon = nullptr;
+    const pat::Electron *ref_electron = nullptr;
+    for(const pat::Muon& muon : muons) { // Look for a muon tag first
+        if(!(muon.pt() > 26 && std::abs(muon.eta()) < 2.1 && muon.isMediumMuon() && PFRelIsolation(muon) < 0.1
```
kandrosov, line 192: - pt > 28 GeV (to be the same for all eras)
- pf_rel_iso < 0.15 (which corresponds to the definition of the tight WP)

Production/src/Selectors.cc
```diff
@@ -177,5 +179,123 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
...
+                && std::abs(muon.muonBestTrack()->dz(primaryVertex.position())) < 0.2
+                && std::abs(muon.muonBestTrack()->dxy(primaryVertex.position())) < 0.045))
+            continue;
+        if(!ref_muon || PFRelIsolation(*ref_muon) > PFRelIsolation(muon) || (PFRelIsolation(*ref_muon) == PFRelIsolation(muon) && ref_muon->pt() < muon.pt())){
+            ref_muon = &muon;
+        }
+    }
+    if(!ref_muon) {
+    	for(const pat::Electron& electron : electrons) { // Start looking for an ele tag if no muon was found
+            if(!(electron.pt() > 35 && std::abs(electron.eta()) < 2.3 && electron.electronID("mvaEleID-Fall17-noIso-V2-wp90") > 0.5f
```
kandrosov, line 202: - pt > 34 GeV
- add pf_rel_iso < 0.15

Production/src/Selectors.cc
```diff
@@ -177,5 +179,123 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
...
+                     && std::abs(electron.bestTrack()->dz(primaryVertex.position())) < 0.2
+                     && std::abs(electron.bestTrack()->dxy(primaryVertex.position())) < 0.045))
+            	 continue;
+            if(!ref_electron || ref_electron->pt() < electron.pt()){
```
kandrosov, line 206: choose the most isolated (same as for ref_muon)

Production/src/Selectors.cc
```diff
@@ -177,5 +179,123 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
...
+            if(!tauJet.jet) continue;
+	    const pat::Jet& jet = *tauJet.jet;
+            if(!(std::abs(jet.eta()) < 2.8 && reco::deltaR(jet.polarP4(), ref_muon->polarP4()) > 0.8))
+		    continue;
+	    if(!selected || selected->jet->pt() < jet.pt()) {
+		    selected = &tauJet;
+		    selectedTauJets = { selected };
+            }
+	}
+	return Result(selectedTauJets, tagObject);
```
kandrosov, line 235: Unify code to select tauJet for ref_muon, ref_electron and "ref_MET" cases to avoid duplications.

Production/src/Selectors.cc
```diff
@@ -177,5 +179,123 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
...
+        for(const TauJet& tauJet : tauJets) {
+            if(!(tauJet.genLepton && tauJet.genJet) && !(std::abs(tauJet.genLepton->visibleP4().eta()) < 2.8 || std::abs(tauJet.genJet->eta()) < 2.8)) continue;
+	    selectedTauJets.push_back(&tauJet); //Fill with jets passing selection
+	}
+	if(selectedTauJets.size() <= 2)
+	    return Result(selectedTauJets, nullptr); //Send result if vector size is already <= 2
+	else { // Otherwise order by genLepton properties
+            std::vector<const TauJet*> newSelectedTauJets;
+	    std::vector<int> genLeptonKind{5,3,4,2,1,6}; // Ordered by lepton kind priority, couldn't come up with a more efficient idea for now...
+	    std::sort(selectedTauJets.begin(), selectedTauJets.end(), [](auto &i, auto &j){return i->genLepton->visibleP4().pt() < j->genLepton->visibleP4().pt();});
```
kandrosov, line 286: - no need to treat size < 2 separately. After the vector is sorted it is enough to call `selectedTauJets.resize(std::min(selectedTauJets.size(), 2))`

I propose following modification: don't include kind priority also for di-tau case - save two jets with the highest pt. However, for all cases (with tag/MET/ditau) in addition to the selected tauJets add all tauJets that have genJepton.kind() == 5 (i.e. true taus). This would allow us
- unbiased estimation of tau reco efficiency
- unbiased selection of the jet pair for di-tau case that is the same for data&mc => also ditau can be used for domain adaptation.

kandrosov (changes requested):

Production/src/Selectors.cc
msessini: So if I understand correctly we would have :

- 1 reco jet with highest pT for tag/MET as we did before + all genLeptons with kind == 5 
- 2 genLeptons with highest pT for ditau regardless of their kind + all genLeptons with kind == 5 instead of two genLeptons sorted by kind priority

Production/src/Selectors.cc
```diff
@@ -177,5 +179,94 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
...
+		tagObject->kind = tau_analysis::selectors::TagObject::Kind::Electron; //new tag object member
+        	tagObject->p4 = ref_electron->polarP4();
+        	tagObject->charge = ref_electron->charge();
+        	tagObject->id = ref_electron->electronID("mvaEleID-Fall17-noIso-V2-wp90"); //Not sure what to put here 
+        	tagObject->isolation = PFRelIsolation(*ref_electron,rho);
+        	tagObject->has_extramuon = hasExtraMuon(muons, ref_muon, primaryVertex);
+        	tagObject->has_extraelectron = hasExtraElectron(electrons, rho);
+        	tagObject->has_dimuon = hasExtraDimuon(muons, primaryVertex);	
+    	}
+    	else { // high MET case
```
shedprog, line 237: It would be nice to have some flag "which of the selection group it is" (1), (2), (3) or (4) since now tagObject kind is only valid for muon/electron tag

Production/src/Selectors.cc
```diff
@@ -177,5 +179,94 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
...
+        selectedTauJets.resize(std::min(static_cast<int>(selectedTauJets.size()), 2));
+    }
+   
+    // Add true taus to selectedTauJets for all cases
+    for(const TauJet& tauJet : tauJets) {
+       if(!(tauJet.genLepton) || !(std::abs(tauJet.genLepton->visibleP4().eta()) < 2.8 && tauJet.genLepton->kind() == reco_tau::gen_truth::GenLepton::Kind::TauDecayedToHadrons)) continue;
+         selectedTauJets.push_back(&tauJet); //Fill with jets passing selection
+    }
+    //
+    return Result(selectedTauJets, tagObject);
```
shedprog, line 268: In the following part there will be no prioritization to the genLepton->kind (no `kind: tau_h > tau_e > tau_mu > e > mu > jet`) do we want to have it?

Also: the following part does require having >2 valid TauJets, from Konstantin's slides I had understanding it was needed.

Production/src/Selectors.cc
shedprog: I see, but in this case there might be confusion, since: tag Object is filled even though MET or gen lepton might not be a tag object in general sense

kandrosov (changes requested):

Production/src/Selectors.cc
```diff
@@ -177,5 +179,94 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
...
+    	}
+    	//	
+    	else if(ref_electron) { // Fill tag with ele if ele found rather than muon
+		tagObject->kind = tau_analysis::selectors::TagObject::Kind::Electron; //new tag object member
+        	tagObject->p4 = ref_electron->polarP4();
+        	tagObject->charge = ref_electron->charge();
+        	tagObject->id = ref_electron->electronID("mvaEleID-Fall17-noIso-V2-wp90"); //Not sure what to put here 
+        	tagObject->isolation = PFRelIsolation(*ref_electron,rho);
+        	tagObject->has_extramuon = hasExtraMuon(muons, ref_muon, primaryVertex);
+        	tagObject->has_extraelectron = hasExtraElectron(electrons, rho);
```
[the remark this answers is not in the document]
msessini: I took the function as it was, do you want me to modify it ? 

Also makes me think that I have seen ```std::abs(muon.muonBestTrack()->dxy(primaryVertex.position())) < 0.0045)``` in several places of the code. Is the unit different for this one or shouldn't it be 0.045 ?

Production/src/Selectors.cc
```diff
@@ -177,5 +179,94 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
...
+            const pat::Jet& jet = *tauJet.jet;
+            if(!(std::abs(jet.eta()) < 2.8) || ((ref_muon || ref_electron) && !(reco::deltaR(jet.p4(), tagObject->p4) > 0.8)))
+                    continue;
+            if(!selected || selected->jet->pt() < jet.pt()) {
+                    selected = &tauJet;
+                    selectedTauJets = { selected };
+            }
+        }
+    }
+    else { // No mu or ele tag nor high MET
```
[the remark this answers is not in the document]
msessini: Yes ```met.pt() > 180``` is included in the corresponding if loop, but I think that is what we want or did I miss something ?

Production/src/Selectors.cc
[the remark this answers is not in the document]
kandrosov: - yes, the old function was only working for muTau channel => you need to modify it.
- should be 0.045 everywhere. 0.0045 is a bug.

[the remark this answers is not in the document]
msessini: Done, should we have a di-electron flag as well by the way ?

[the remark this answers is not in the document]
kandrosov: yes, it can also be added. Although, I'm not sure if we going to use these di-muon or di-electron flags..

Analysis/interface/TauTuple.h
```diff
@@ -145,6 +145,9 @@
     VAR(Int_t, has_extramuon) /* Extra muon present */ \
     VAR(Int_t, has_extraelectron) /* Extra electron present */ \
     VAR(Int_t, has_dimuon) /* Extra muon pair present */ \
+    VAR(Int_t, has_dielectron) /* Extra electron pair present */ \
+    /* Selection type of TagAndProbe selector */ \
+    VAR (Int_t, selectionType) /* 1 Muon tag, 2 Ele tag, 3 High MET, 4 gen based */ \
```
kandrosov, line 150:
```suggestion
    VAR (Int_t, selectionType) /* 1 Muon tag, 2 Ele tag, 3 High MET, 4 pt order, 5 gen based */ \
```

Production/src/Selectors.cc
```diff
@@ -251,6 +272,7 @@ TauJetSelector::Result TagAndProbe::Select(const edm::Event& event, const std::d
     }
     else { // No mu or ele tag nor high MET 
         tagObject = nullptr;
+        TauJetSelector::selectionType = tau_analysis::selectors::TauJetSelector::Type::GenBased;
```
kandrosov, line 275: selection type is property of a tau. E.g. in the same event there could be one tau selected by t&p, while few other taus are selected because they are true tau_h

kandrosov (changes requested):

Analysis/interface/TauTuple.h
```diff
@@ -162,6 +164,8 @@
     TAU_VAR(Int_t, index) /* index of the tau */ \
     TAU_VAR4(Float_t, pt, eta, phi, mass) /* 4-momentum of the tau */ \
     TAU_VAR(Int_t, charge) /* tau charge */ \
+    /* Tau selection type of TagAndProbe selector */ \
+    TAU_VAR (Int_t, selectionType) /* 1 Muon tag, 2 Ele tag, 3 High MET, 4 Pt ordered, 5 gen based */ \
```
kandrosov, line 168: it is not `TAU_VAR`, but rather `VAR` as it was before. It could be that HPS tau is not reconstructed.

Production/plugins/TauTupleProducer.cc
```diff
@@ -271,11 +272,13 @@ class TauTupleProducer : public edm::one::EDAnalyzer<> {
         tauTuple().has_extramuon = tagObj ? tagObj->has_extramuon : default_value;
         tauTuple().has_extraelectron = tagObj ? tagObj->has_extraelectron : default_value;
         tauTuple().has_dimuon = tagObj ? tagObj->has_dimuon : default_value;
+        tauTuple().has_dielectron = tagObj ? tagObj->has_dielectron : default_value;
 
         tauTuple().total_entries = static_cast<int>(tauJets.size());
         for(size_t tauJetIndex = 0; tauJetIndex < tauJets.size(); ++tauJetIndex) {
             const TauJet& tauJet = *tauJets.at(tauJetIndex);
             tauTuple().entry_index = static_cast<int>(tauJetIndex);
+	    tauTuple().tau_selectionType = static_cast<int>(selectedTypes.at(static_cast<int>(tauJetIndex)));
```
kandrosov, line 281:
```suggestion
	    tauTuple().tau_selectionType = tauJetIndex < selectedTypes.size() ? static_cast<int>(selectedTypes.at(static_cast<int>(tauJetIndex))) : -1;
```

Production/src/Selectors.cc
```diff
@@ -174,7 +192,104 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
...
+        }
+    }
+    else { // No mu or ele tag nor high MET 
+        tagObject = nullptr;
+        for(const TauJet& tauJet : tauJets) {
+            if(!(tauJet.genLepton) || !(std::abs(tauJet.genLepton->visibleP4().eta()) < 2.8) || !(tauJet.genJet && std::abs(tauJet.genJet->eta()) < 2.8)) continue;
+	    selectedTauJets.push_back(&tauJet); //Fill with jets passing selection
+            selectedTypes.push_back(tau_analysis::selectors::TauJetSelector::Type::PtOrdered);
+	}
+        std::sort(selectedTauJets.begin(), selectedTauJets.end(), [](auto &i, auto &j){return i->genLepton->visibleP4().pt() < j->genLepton->visibleP4().pt();});
```
kandrosov, line 281:
```suggestion
        std::sort(selectedTauJets.begin(), selectedTauJets.end(), [](auto &i, auto &j){return i->genLepton->visibleP4().pt() > j->genLepton->visibleP4().pt();});
```

Production/src/Selectors.cc
```diff
@@ -174,7 +192,104 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
...
+        selectedTauJets.resize(std::min(static_cast<int>(selectedTauJets.size()), 2));
```
kandrosov, line 282: also resize `selectedTypes`

Production/src/Selectors.cc
```diff
@@ -174,7 +192,104 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
...
+		tagObject->has_dielectron = hasExtraDielectron(electrons, rho);
+    	}
+    	else { // high MET case
+        	tagObject = nullptr;   
+    	}
+        const TauJet* selected = nullptr;
+        for(const TauJet& tauJet : tauJets) { //Look for highest pT jet
+            if(!tauJet.jet) continue;
+            const pat::Jet& jet = *tauJet.jet;
+            if(!(std::abs(jet.eta()) < 2.8) || ((ref_muon || ref_electron) && !(reco::deltaR(jet.p4(), tagObject->p4) > 0.8)))
```
kandrosov, line 263: I was thinking about deltaR requirement. Maybe 0.8 is too much: current deepTau training still show good performance for small deltaR, so I think it is safe to go to 0.4. Would you agree?

Production/src/Selectors.cc
```diff
@@ -174,7 +192,104 @@ TauJetSelector::Result TauJetTag::Select(const edm::Event& event, const std::deq
...
+	    selectedTauJets.push_back(&tauJet); //Fill with jets passing selection
+            selectedTypes.push_back(tau_analysis::selectors::TauJetSelector::Type::PtOrdered);
+	}
+        std::sort(selectedTauJets.begin(), selectedTauJets.end(), [](auto &i, auto &j){return i->genLepton->visibleP4().pt() < j->genLepton->visibleP4().pt();});
+        selectedTauJets.resize(std::min(static_cast<int>(selectedTauJets.size()), 2));
+    }
+   
+    // Add true taus to selectedTauJets for all cases
+    for(const TauJet& tauJet : tauJets) {
+       if(!(tauJet.genLepton) || !(std::abs(tauJet.genLepton->visibleP4().eta()) < 2.8 && tauJet.genLepton->kind() == reco_tau::gen_truth::GenLepton::Kind::TauDecayedToHadrons) || (selectedTauJets.size()!=0 && &tauJet == selectedTauJets.at(0))) continue;
```
kandrosov, line 287: you need to loop: size of selecteTauJets could be 2 (in case of no tag or met)

kandrosov (changes requested):

kandrosov (approved):

